PR #15, opened 2017-05-15
Add German translation

Strayer:
Hi! I translated the extension to German.

Please check my additional line comments, I changed some things and have some suggestions.

### DIFF
--- a/src/_locales/de/messages.json
+++ b/src/_locales/de/messages.json
@@ -45,7 +45,7 @@
   },
 
   "clearCookiesForDomainText": {
-    "message": "Lösche alle Cookies dieser Domain",
+    "message": "Alle Cookies dieser Domain löschen",
     "description": "Clear All Cookies For This Domain"
   },
 
@@ -105,7 +105,7 @@
   },
 
 "activeModeText": {
-    "message": "Aktiven Modus aktivieren? Verzögerung bis zum Aufräumen:",
+    "message": "Aktiven Modus einschalten? Verzögerung bis zum Aufräumen:",
     "description": "Enable Active Mode? Delay Before Cleaning:"
   },
 
@@ -115,7 +115,7 @@
   },
 
 "activeModeTooltipText": {
-    "message": "Wenn der aktive Modus aktiviert ist wird das Aufräumen der Cookies wird beim Schließen eines Tabs ausgeführt.",
+    "message": "Wenn der aktive Modus eingeschaltet ist wird das Aufräumen der Cookies beim Schließen eines Tabs ausgeführt",
     "description": "When active mode is enabled, cookie cleanup is triggered on tab close."
   },
 
@@ -125,7 +125,7 @@
   },
 
 "statLoggingTooltipText": {
-    "message": "Zählt die Anzahl der gelöschten Cookies, wird im Willkommensbildschirm angezeigt.",
+    "message": "Zählt die Anzahl der gelöschten Cookies, wird im Willkommensbildschirm angezeigt",
     "description": "Counts the number of cookies deleted during this session and in total. This is shown in the Welcome Screen."
   },
 
@@ -225,7 +225,7 @@
   },
 
 "rightClickToSave": {
-    "message": "Right Click to Save Link As",
+    "message": "Klicke mit der rechten Maustaste auf diesen Link und wähle \"Ziel speichern unter...\"",
     "description": "Right Click to Save Link As"
   },
 
--- a/src/_locales/de/messages.json
+++ b/src/_locales/de/messages.json
@@ -1,0 +1,254 @@
+{
+  "extensionName": {
+    "message": "Cookie AutoDelete",
+    "description": "Name of the extension."
+  },
+
+  "extensionDescription": {
+    "message": "Beherrsche deine Cookies! Lasse unbenutzte Cookies von deinen geschlossenen Tabs automatisch löschen.",
+    "description": "Control Your Cookies! Auto Delete Unused Cookies From your Closed Tabs While Keeping The Ones You Want."
+  },
+
+  "notificationTitle": {
+    "message": "Cookie AutoDelete: Cookies wurden gelöscht!",
+    "description": "Cookie AutoDelete: Cookies were Deleted!"
+  },
+
+  "notificationContent": {
+    "message": "$Num$ Cookies gelöscht von: $Websites$",
+    "description": "123 Deleted Cookies from: google.com, facebook.com",
+    "placeholders": {
+      "num" : {
+        "content" : "$1",
+        "example" : "123"
+      },
+      "websites" : {
+        "content" : "$2",
+        "example" : "google.com, facebook.com"
+      }
+    }
+  },
+
+  "hostWebsiteText": {
+    "message": "Webseite:",
+    "description": "Host Website:"
+  },
+
+  "cookieCleanupText": {
+    "message": "Cookies aufräumen",
+    "description": "Cookie Cleanup"
+  },
+
+  "cookieCleanupIgnoreOpenTabsText": {
+    "message": "Cookies aufräumen (offene Tabs ignorieren)",
+    "description": "Cookie Cleanup (Ignore Open Tabs)"
+  },
+
+  "clearCookiesForDomainText": {
+    "message": "Lösche alle Cookies dieser Domain",
+    "description": "Clear All Cookies For This Domain"
+  },
+
+  "activeModeSwitchText": {
+    "message": "Aktiver Modus",
+    "description": "Active Mode"
+  },
+
+  "switchToWhiteListText": {
+    "message": "Zur Whitelist hinzufügen?",
+    "description": "Add to WhiteList?"
+  },
+
+  "welcomeText": {
+    "message": "Willkommen",
+    "description": "Welcome"
+  },
+
+  "settingsText": {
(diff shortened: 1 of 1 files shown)


src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
+{
+  "extensionName": {
+    "message": "Cookie AutoDelete",
+    "description": "Name of the extension."
+  },
+
+  "extensionDescription": {
+    "message": "Beherrsche deine Cookies! Lasse unbenutzte Cookies von deinen geschlossenen Tabs automatisch löschen.",
```
Strayer: I changed the text a little bit to say something like "Let unused cookies be automatically deleted" instead of "Auto Delete Unused Cookies". Sounds better in German in my opinion, though I can change this to the original text.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+      "websites" : {
+        "content" : "$2",
+        "example" : "google.com, facebook.com"
+      }
+    }
+  },
+
+  "hostWebsiteText": {
+    "message": "Webseite:",
+    "description": "Host Website:"
```
Strayer: I'm a little bit confused by the original text. What is the meaning of "Host" in this context? As far as I understand this simply shows the domain name of the website I'm currently on… thats why I simply put "website" in german in there, not sure how to translate this.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+  },
+
+  "cookieCleanupIgnoreOpenTabsText": {
+    "message": "Cookies aufräumen (offene Tabs ignorieren)",
+    "description": "Cookie Cleanup (Ignore Open Tabs)"
+  },
+
+  "clearCookiesForDomainText": {
+    "message": "Lösche alle Cookies dieser Domain",
+    "description": "Clear All Cookies For This Domain"
```
Strayer: Just to be sure: This deletes ALL cookies of the domain, even if it is whitelisted, right? "Clear" is a little unclear in this context (heh), since it sounds so similar to "clean".

src/_locales/de/messages.json
mrdokenny: That's fine.

src/_locales/de/messages.json
mrdokenny: Basically what I mean is a website can be something like: https://github.com/mrdokenny/Cookie-AutoDelete/
but the Host Website is github.com

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+  },
+
+  "activeModeSwitchText": {
+    "message": "Aktiver Modus",
+    "description": "Active Mode"
+  },
+
+  "switchToWhiteListText": {
+    "message": "Zur Whitelist hinzufügen?",
+    "description": "Add to WhiteList?"
```
Strayer: Just a nitpick, but I don't like the wording of this… I think "Whitelisted" or something like this makes more sense, since it is a checkbox, no reason for a question. Translated it literally though to stay true to the english source.

src/_locales/de/messages.json
mrdokenny: Yes.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+    "description": "Cookie WhiteList"
+  },
+
+  "aboutText": {
+    "message": "Über",
+    "description": "About"
+  },
+
+  "welcomeMessage": {
+    "message": "Cookie AutoDelete hat während dieser Sitzung $sessionDeleted$ und insgesamt $totalDeleted$ Cookies gelöscht.",
```
Strayer: I skipped the "Hi there" here, since there is already a welcome in the title. Can change this if you want.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+      },
+      "totalDeleted" : {
+        "content" : "$2",
+        "example" : "456"
+      }
+    }
+  },
+
+  "reviewLinkMessage": {
+    "message": "Danke für das Ausprobieren von Cookie AutoDelete. Falls es dir gefällt, verfasse bitte einen Erfahrungsbericht.",
```
Strayer: I chose the informal "Du" instead of the formal "Sie". I think its fine for this extension, though others might disapprove.

src/_locales/de/messages.json
Strayer: Ahh, I see… I guess "domain" would be fitting in this context, not entirely sure. I'll try to check how other projects name this.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+    "description": "Counts the number of cookies deleted during this session and in total. This is shown in the Welcome Screen."
+  },
+
+"resetCounterText": {
+    "message": "Zähler zurücksetzen",
+    "description": "Reset Counter"
+  },
+
+"showNumberOfCookiesInIconText": {
+    "message": "Zeigt die Anzahl der Cookies einer Domain in der Symbolleiste",
```
Strayer: I added "in the toolbar" here, also "in the icons tooltip in the toolbar" in the next translation string. Otherwise the text doesn't add anything helpful for the user, in my opinion.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+  },
+
+"cancelText": {
+    "message": "Abbrechen",
+    "description": "Cancel"
+  },
+
+"defaultSettingsText": {
+    "message": "Standardeinstellungen",
+    "description": "Default Settings"
```
Strayer: Usually I'd suggest "Restore default settings" (and "Standardeinstellungen wiederherstellen") here, but I didn't want to change the meaning too much.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+  },
+
+"defaultSettingsTooltipText": {
+    "message": "WARNUNG: Diese Aktion wird auch deine URLs zurücksetzen. Erstelle eine Sicherung!",
+    "description": "WARNING: This will also clear your URLS as well. So make a backup of them!"
```
Strayer: I think the term "URL" is confusing here, since this is talking about the whitelist, as far as I understand. Maybe this should be changed to "WARNING: This will clear your whitelist as well, make sure you have a backup!"?

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+  },
+
+"enterURLTooltipText": {
+    "message": "Trage URLs ohne 'www.' ein, drücke dann die Eingabetaste oder klicke auf 'URL hinzufügen'. Alternativ besuche die Seite und füge sie vom blauen Symbol in der Symbolleiste hinzu.",
+    "description": "Enter URLS without the 'www.' part, then press 'Enter' or 'Add URL'. Alternatively go to the site and add it from the blue icon."
+  },
+
+"exportURLSText": {
+    "message": "URLs exportieren",
+    "description": "Export URLS"
```
Strayer: As above, I'd suggest to say something like "Export whitelist" here.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+  },
+
+"rightClickToSave": {
+    "message": "Klicke mit der rechten Maustaste auf diesen Link und wähle \"Ziel speichern unter...\"",
+    "description": "Right Click to Save Link As"
+  },
+
+"importURLSText": {
+    "message": "Datei hochladen",
+    "description": "Upload File"
```
Strayer: To keep consistency I'd suggest changing this to "Restore the whitelist from an exported backup file" or something like that… a native speaker could probably give a better suggestion. Translated it literally for now.

src/_locales/de/messages.json
```diff
@@ -0,0 +1,254 @@
...
+  },
+
+"reportIssuesText": {
+    "message": "Melde Fehler und schlage neue Funktionen vor",
+    "description": "Report issues and suggest features"
+  },  
+
+"versionText": {
+    "message": "version",
+    "description": "version"
```
Strayer: Now this is tricky… as far as I can see this is used in the about screen for "Chrome version" and "Firefox version". The problem is that you would express this as "Chrome-Version" and "Firefox-Version" in German. Could be changed to a format string.

Strayer:
Sorry, adding the line comments took longer than I expected… :) Done now.

mrdokenny:
@Strayer Still waiting on "Could you explain what's the usage for "Contextual Identities" since the original did not have them?"

Strayer:
@mrdokenny Not sure what you mean. If you're talking about [line 158](https://github.com/mrdokenny/Cookie-AutoDelete/pull/15/files#diff-869200a8548d888da8a18799c8ee9ac3R158): I took that from the English translation file and it seems to be used in the settings screen. If you mean why I put that into quotes, its because I was not sure how to translate this properly and decided to keep the English description of the container context.

Did I maybe miss a comment you made in the diff? Sorry for the confusion…

mrdokenny:
> If you mean why I put that into quotes, its because I was not sure how to translate this properly and decided to keep the English description of the container context.

@Strayer Yeah that's pretty much what I was wondering. I'll merge this in now and take your English suggestions into consideration. Thanks for your contribution!

PR #19, opened 2017-05-17
Create french messages.json

Primokorn:
Not sure if we have to translate 'Entrer' and 'Add URL' in "enterURLTooltipText".


mrdokenny:
'Enter' meaning the Enter key on the keyboard and 'Add URL' is the name of the button of "addURLText"

Primokorn:
Yes, I know but sometimes we have keywords in a description which don't need a translation since they are like references.

PR #24, opened 2017-05-19
Update bulgarian translation

AdmiralAnimE:
:)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ jspm_packages
 .node_repl_history
 *.xpi
 *.zip
+src/icons/Thumbs.db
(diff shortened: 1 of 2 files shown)


AdmiralAnimE:
I don't think it's necessary on those lines. There is no point in repeating the name like in english, to me its perfectly understandable from the context.

AdmiralAnimE:
One request - can you make line 243 to be "$1 version" where $1 is variable, because I want to write "version for $" ?

mrdokenny:
> One request - can you make line 243 to be "$1 version" where $1 is variable, because I want to write "version for $1" ?

That's something that can be done on the next release. 

Also there's some edits I made to the wording of some strings as well as added 2 new ones that you might want to check out.

PR #55, opened 2017-06-20
Update pt_PT

SW1FT:


mrdokenny:
https://www.diffchecker.com/k1RDOx8n

@SW1FT Thanks but you are missing two translations according to the diff tool on line 57.

mrdokenny:
Should be good, thanks!

PR #56, opened 2017-06-21
Update FR messages.json

Primokorn:
As requested.


mrdokenny:
There's still this one 
` "enableGlobalSubdomainTooltipText": {
 "message": "When enabled, *.example.com will match example.com. Turning this off gives you more granular control over your lists.",
"description": "When enabled, *.example.com will match example.com. Turning this off gives you more granular control over your lists."		
},`

This is after "enableGlobalSubdomainText" which is line 182 on this [diff ](https://www.diffchecker.com/IuyBytK5)

Primokorn:
That's weird. I'm sure that I already translated this string...

Issue #57, opened 2017-06-23
1.4.0 Before Release Todos


mrdokenny:
I just released 1.4.0b3 to the AMO development channel for wider audience testing. The plan is to wait for about a week before fully releasing on stable channels. In the meantime, no new functionality will be added to 1.4.0, only bug fixes and the below todos. Report any bugs by making a new issue or commenting on the feature tracking issues.

- [x] Visual tweaks via CSS (some parts of UI look off-center)
- [x] Update locales
- [x] Update Screenshots
- [x] Update Descriptions

@Strayer @yfdyh000 @AdmiralAnimE  
There's been a lot of changes to the `extension/_locales/en/manifest.json` file including moving it to a different folder which caused some updates to be lost. I recommend using a diff tool like this [(sample de locale diff with en)](https://www.diffchecker.com/IkRxgSXg) to check what has been changed. 

Basically, make a new fork, copy any missing strings, and update the current strings if the "description" part is different in the locales in any of the strings (also remember to update the description part as well).

In the future, I'll probably move to Crowdin once CAD becomes feature complete and there won't be as many strings to update.

Thanks to all the translators, testers, bug reporters, etc.

Strayer:
Updated German translation in PR #63

PR #58, opened 2017-06-23
Update zh_CN to 1.4.0

yfdyh000:
#57. Not tested, I encountered an error when `npm run-script build`:
```
...
19 verbose node v8.1.2
20 verbose npm  v5.0.3
21 error code ELIFECYCLE
22 error errno 1
23 error cookie-autodelete@1.4.0 build: `webpack --config webpack.config.js && ./tools/buildFilesDev.sh`
```
Due to I using Windows 10.

### DIFF
--- a/extension/_locales/zh_CN/messages.json
+++ b/extension/_locales/zh_CN/messages.json
@@ -54,9 +54,19 @@
     "description": "Active Mode"
   },
 
+  "switchToNoListText": {
+    "message": "标签页关闭后清理",
+    "description": "Clean After Tab Close"
+  },  
+
+  "switchToGreyListText": {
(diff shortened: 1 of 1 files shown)


mrdokenny:
>Due to I using Windows 10.

Oh, it works on bash linux not cmd, that's probably why. On windows, I use [Bash for Windows](https://www.howtogeek.com/249966/how-to-install-and-use-the-linux-bash-shell-on-windows-10/) and found it to work quite well. Since it's similar to bash on linux, you can install node.js using the package manager.

mrdokenny:
@yfdyh000 Here's what it looks like on mine.
screenshot 2017-06-22 20 27 59

PR #90, opened 2017-08-01
Added Serbian translation

kostich:


mrdokenny:
I'm going to change it manually. Thanks anyway.

kostich:
Ah, so you would like for 'Cookie AutoDelete' to be left as is? Like what's done with the trademarks?

mrdokenny:
@kostich Yes 😉

kostich:
Ok, will take that in account the next time I update the translations. :)

PR #91, opened 2017-08-04
Refine translation

StoyanDimitrov:
Change few terms to align better with terms in the browser UI, quotes are changed too.


mrdokenny:
Thanks for your contribution.

PR #133, opened 2017-09-03
wrap setting txts in labels

chpio:
Labels connect a text with a form element, allowing to activate the form element by clicking on the text.
https://developer.mozilla.org/en/docs/Web/HTML/Element/label


mrdokenny:
I can merge this and probably release one more point release of 1.4.4 but just to let you know that I'm now focusing on 2.X.X (redone in React/Redux) which I already included these changes. Thanks anyway.

PR #161, opened 2017-09-26
Inline editable table

seansfkelley:
(Built on top of #160 so that should merge first.)

This makes the table much more compact and the editing interactions physically closer to the things they interact with (rather than having a full-on "switch to edit mode" button):

On hover, show an "edit" button by the expression for editing the expression:

image

Show a live preview of the regex when editing the expression:

image

One-click toggling between whitelist/graylist:

image


ke-d:
This looks amazing! When I have time, I'll take a closer look at this before merging. Thanks!

test/CleanupService-spec.js
```diff
@@ -10,21 +10,21 @@ const sampleState = {
 	"lists": {
 		"default": [
 			{
-				"regExp": "^.*google\\.com$",
```
seansfkelley, line 13: I'm not totally sure that this is the right way to fix these tests, but I _think_ it is?

seansfkelley:
Thanks! I noticed space for these improvements when setting it up for myself and others and wanted to contribute. I have a couple other ideas around the settings page and reordering the popup that I'm also working on. Thanks for making a good replacement to Self-Destructing Cookies, as well!

PR #163, opened 2017-09-27
Restructure popup

seansfkelley:
(This is built on top of #161 so that should merge first.)

I went a little crazy restructuring the popup to make it simpler while clarifying what actions are possible. Here's what these changes look like (with an example tooltip visible):

image

I'm open to discussing details of the visuals/arrangement, but I think this is a big improvement in comprehensibility already. Changes:

- reduce redundancy in the popup (replacing words with icons, restructuring the domain-specific section)
- improve visual cues (more colors and icons)
- improve visual consistency (more icons, more consistent sizes and spacing)
- add tooltips to icon buttons/one-word buttons
- tweak wording of related tooltips/buttons
- reorder the elements so that all the domain-specific stuff is at the bottom, together
- don't show the table if there are no matches

On mobile, it should reflow to fit as follows, though I don't have a proper phone to verify on. It's not the best but I've never actually used Bootstrap before so it's possible I haven't structured the DOM quite right for it.

image


ke-d:
Yeah, this looks better than what I can come up work. Thanks.

A couple of thoughts:
A couple of people suggested `Enable/Disable Automatic Cleaning` in https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/issues/104#issuecomment-327381746. 
I might change the phrasing on the `Ignore open tabs` because of https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/issues/136#issuecomment-327418826

ghost:
How this look on mobile?

seansfkelley:
@mrdokenny thanks! On the phrasing changes:

I changed the wording of the enable/disable to (I think) be clearer. It also has a tooltip now; check the screenshot in the original description.

Re: "ignore", I agree that it's confusing. I might change the phrasings to "clean" and "clean, preserving open tabs" and then have explanatory hovers, perhaps (respectively) "clean cookies, including those for open tabs" and "clean cookies, preserving those for open tabs".

I was also toying with the idea of using a dropdown for the second option, and then reversing the intent of the buttons such that the less destructive option is the simplest/more immediate, i.e., "clean, preserving open tabs", and the more destructive option was a little hidden. I didn't pursue it, and would prefer not to block this PR on it, since this PR doesn't change that. But, for your consideration, it might look like the following.

By default, showing hover for the "primary action":

image

When opening the dropdown to get the more destructive option:

image

(This is just a mockup so the alignment is a bit wonky.)

There are some problems around this interaction due to mobile support (though I'm unclear if Cookie AutoDelete actually supports mobile?) and the fact that dropdowns inside a popup are kind of silly.

ghost:
>though I'm unclear if Cookie AutoDelete actually supports mobile?

I have stable version installed on nightly. To be honest - I imported whitelist from desktop just now, but it seems to work. Lot of cookies has been deleted - I get notification in Android status area. This may be too distracting - I will probably disable notifications. Popup (actually new tab) can be opened from main menu - it's not looking too good right now.

seansfkelley:
@gwarser I don't have a phone to verify mobile support on, so I did my best. If you want to pull these changes and try them out, please do, but all I can really manage is making my Firefox window very small and seeing how it behaves (I can't even use proper mobile emulation from Firefox due to a weird restriction Firefox imposes for extensions' pages...).

B00ze64:
@seansfkelley Nice work. Your 2nd screenshot shows no space between the buttons, unlike the first screenshot at the start of this bug. I *much* prefer with some space between each of the buttons. In fact I think there should be a space between the 2 orange "clean" buttons. The color already helps group them together, so we already know they are related. In my mind buttons should be clearly separated, otherwise it just looks like a big bar and it's not clear we can actually click on its different colors; it's not clear they are buttons...

seansfkelley:
@B00ze64 good call; I spaced the buttons out and also updated my comments to be a little less ambiguous about what it actually looks like now.

B00ze64:
@seansfkelley Thanks! Man this is going to look so good! Some people really have a nack for UI ;-)

seansfkelley:
Thanks @mrdokenny!

ke-d:
@seansfkelley Thank you for making the UI better and aesthetically pleasing. :)

grenzor:
IMO I like the idea of the dropdown. Something like:

```
Clean Cookies
>Clean Cookies (Including Open Tabs)
>Clean Cookies (For the Current Domain only)
```

Currently, these 3 options do seem a bit repetitive with 3 separate buttons and it makes the popup a little larger/cluttered than need be.

PR #173, opened 2017-10-08
Update for German locale

spodermenpls:


ke-d:
I will upload all the locales from Crowdin before the next release, so I won't be taking anymore locale PRs.

Issue #178, opened 2017-10-13
Weird UI when put into drop down list


waqas:
right-click on toolbar -> Customise...
Like this...
![ ](https://user-images.githubusercontent.com/407/31532137-e02b0204-b004-11e7-858c-1e5b53a4c6be.png)

ke-d:
It looks like the overflow menu doesn't like a min-width for the popup?

waqas:
Apprently, yes. Previous popup UI was also behaving like this.

ke-d:
I'm not sure if I have to do anything or if Mozilla has to fix anything on their end?

I also noticed a couple of UI problems in the overflow menu with other extensions as well.

@jonathanKingston Do you know if there's a Bugzilla bug for this?

grenzor:
I believe this is the bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1373490

uniquePWD:
It feels like the new UI was designed in a desktop bubble. It seems to stretch horizontally when both on desktop and mobile, it makes more sense to design it vertically. I hope that @seansfkelley can resubmit a better and more useful design. Also it was far more useful (to me at least) to show the list of cleared cookies than just show the rules for the current domain in the popup.

Anyway, here's another screenshot to show how broken it looks on desktop
![cad-newui-sucks](https://user-images.githubusercontent.com/203217/31557994-164cfd2c-b043-11e7-9fbe-fd40f88c3864.png)

seansfkelley:
@pwd-github just submitted #180 to fix the most egregious problems with the layout. I plan to take a deeper look at it over the next few days now that the design has had some time to bake/be in the wild.

The replacing of the "most recently cleared cookies" note on the bottom with the table was prior to my work, though I personally always found that note to be misleading and noisy and was happy to see it gone.

ke-d:
Note that I added a `min-width` property in 8d3726bc9562876389e14547cebf8568152158b6 to fix the popup size in Chrome. What I could do (if it's still needed after @seansfkelley's PR), is apply it only on Chrome.

uniquePWD:
[...]

Issue #174, opened 2017-10-09
Feedback about new interface (on mobile)


ghost:
### Describe the Bug/Feature

Popup could be wider:

![2017-10-09-13-11-44](https://user-images.githubusercontent.com/886325/31335861-b092e754-acf4-11e7-8e06-b8a987895bef.png)

![2017-10-09-13-11-37](https://user-images.githubusercontent.com/886325/31335891-d27a5c1c-acf4-11e7-9f04-755fda91a28c.png)

On first run after update, Settings -> List of expressions - only first row of the table was visible:

![2017-10-09-13-15-16](https://user-images.githubusercontent.com/886325/31335932-fc14cf08-acf4-11e7-992a-ee3f4989c121.png)

![2017-10-09-13-15-09](https://user-images.githubusercontent.com/886325/31335863-b2d38c8a-acf4-11e7-88b7-cd3b1abe85de.png)

This fixed itself after refresh or two.

### Your settings 

- OS/version: 4.4.2 480x800 DPR 1.5
- Browser/version: Firefox Nightly
- Cookie AutoDelete version: 2.0.0b4

ghost:
Also, not mobile, but small screen:

![screenshot_20171009_134753](https://user-images.githubusercontent.com/886325/31336800-ceee6896-acf8-11e7-9890-8f4c1936c837.png)

Content must be scrolled by width of side menu.

ke-d:
I probably won't do the popup issue until a later release since debugging on mobile is harder than on desktop.

ghost:
@mrdokenny can this side menu be auto-closed on some events? On menu click / mouse out / on page click?
Can last used page be saved?

Now, every time I go to Settings, Release Notes are shown. I must click this relatively small button to open menu, go to Cookie Settings or List of Expressions, and then click again, to do anything.

fichtennadel:
What about adding a "quick 'add to WL' option" in FF menu on Android, like SDC had? 

On mobile it's currently quite many steps to add domains, on a hard to read/navigate UI (elements are to small).

ghost:
Bit small, but usable.

@fichtennadel I propose, you create new issue for your request.

PR #180, opened 2017-10-13
Fix popup reflow.

seansfkelley:
Addresses #178 and some of #174.

This PR updates Bootstrap to version 4 to get flexbox-based grids, which are considerably easier to use and tweak.

The visuals now have slightly different sizes/colors per the Bootstrap update (I kind of like it, the buttons were pretty huge before).

image

When Cookie AutoDelete is stuck into the "more" menu, it now shrinks to fit. It isn't the most amazing visuals in the world, but this was an issue before the UI redesign and I want to get the fix out there sooner rather than later. These changes _should_ also make it more tolerant of being on a mobile device, though as noted on my previous PRs, I don't have a phone to test this on and can only check by noodling my desktop browser's size.

(FF 56 on OS X)

image

I also had to make some slight changes to the buttons on the Expressions settings page to bring them up to speed. I verified there are no behavioral regressions in them w/r/t the commit this PR is forked from.


ghost:
Are you aware of this?
image
Text in "Clean, include open tabs" button may be very long in other languages https://crowdin.com/translate/cookie-autodelete/31/en-pl#930

seansfkelley:
Yeah, it's pretty long already in English. I'm tooling around with a design that's more halfway between this one and the previous design to accommodate lengthy lines/phones better, but I wanted to get this change out there quickly so it at least wasn't totally busted for some users.

ke-d:
I think the upgrade to Bootstrap 4.0 caused this issue with this nav tab for Containers Tab Mode.

Original:
capture1

This PR:
capture2

seansfkelley:
@mrdokenny fixed contextual identities. Are there other bits of UI that don't appear with the default configuration I should check?

ke-d:
@seansfkelley Seems like in #182 and #183 the problem might be the UI is too wide. So maybe combining the two clean functions into one and just using a dropdown like what you proposed here:

image

image

ke-d:
I'll merge this as is for now while I fix a couple of issues.

PR #198, opened 2017-10-17
Replace clean buttons by dropdown.

seansfkelley:
Condense the header and collect all the "clean" operations into one place. Screenshot showing dropdown opened with on-hover tooltip:

image

This should keep the dropdown visible even for narrow screens:

image

I also enlarged the buttons to make them better click targets on mobile/not fight with Bootstrap on styles.

### DIFF
--- a/extension/popup/popup.html
+++ b/extension/popup/popup.html
@@ -14,6 +14,10 @@
     <script src="../global_files/redux.js"></script>
     <script src="../global_files/react-redux.js"></script>
     <script src="../global_files/redux-webext.js"></script>
+    <script src="../global_files/jquery-3.2.1.slim.js"></script>
+    <script src="../global_files/popper.js"></script>
+    <script src="../global_files/bootstrap/util.js"></script>
+    <script src="../global_files/bootstrap/dropdown.js"></script>
     <script src="./popupBundle.js"></script>
   </body>
 </html>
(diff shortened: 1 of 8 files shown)


extension/popup/popup.html
```diff
@@ -14,6 +14,10 @@
     <script src="../global_files/redux.js"></script>
     <script src="../global_files/react-redux.js"></script>
     <script src="../global_files/redux-webext.js"></script>
+    <script src="../global_files/jquery-3.2.1.slim.js"></script>
+    <script src="../global_files/popper.js"></script>
+    <script src="../global_files/bootstrap/util.js"></script>
+    <script src="../global_files/bootstrap/dropdown.js"></script>
```
seansfkelley, line 20: There were three ways to approach dropdowns:

1. Do it myself (no thanks).
2. Use react-bootstrap (tried; only supports Bootstrap 3).
3. Do it how Bootstrap wants it and include jQuery/Popper/etc.

It's a bit of a bummer since mixing jQuery and React can be very bad, but since we don't actually use jQuery directly in the React code, it seems okay: React is always rendering the same elements, Bootstrap-via-jQuery is just noodling some of the styles. We don't even have to do weird stuff with refs!

ke-d:
I'll see if I could make my own version for the bootstrap dropdown. I have some dropdown code in the settings menu of 1.X.X that I can look at.

seansfkelley:
@mrdokenny If you'd like, though this seems to work just fine at the moment.

ke-d:
How much of a file size increase are we talking here?

seansfkelley:
Unminified and uncompressed:

- jQuery: 210k
- Popper: 80k
- dropdown: 14k
- util: 4k

308k total. The extension is currently 2.1m, so it's fairly small overall.

seansfkelley:
(oops)

ke-d:
Hmm, since Mozilla is doing automatic publishing, I can probably switch some libraries to the minified versions to gain back some size.

seansfkelley:
There are both minified and unminified files currently present in the extension, so I figure I'll leave it at your discretion to do it all at once if you so decide.

ke-d:
I only had some of them minified cause Mozilla had hashes for those for when they were doing manual reviews. So those libraries didn't have to be manually checked.

seansfkelley:
Ah, interesting, I didn't know they did that. Makes sense.

seansfkelley:
Ping @mrdokenny anything you'd like me to do or is it good to go?

ke-d:
@seansfkelley I have been busy but I'll take a look at it this weekend. :)

ke-d:
Thanks for the contribution!

PR #206, opened 2017-10-25
Fix #193

### LINKED ISSUE #193: Request: Hide badge when number of cookies is 0
### Describe the Bug/Feature
When I view a page that WebExtensions can't affect (i.e. an about: or moz-extension: page), my addon bar strip looks like this: 

As you can see, Cookie Autodelete sticks out like a sore thumb, being the only one with a badge. I wish the badge was hidden whenever it would display "0." If there is no badge on a page, it'd be safe to assume there are no cookies on the domain.

christianheinrichs:


src/services/BrowserActionService.js
```diff
@@ -17,10 +18,13 @@ export const showNumberOfCookiesInIcon = async (tab) => {
...
-		text: cookies.length.toString(),
-		tabId: tab.id
-	});
+
+	if (cookies.length !== 0) {
+		browser.browserAction.setBadgeText({
+			text: cookies.length.toString(),
+			tabId: tab.id
+		});
+	}
```
seansfkelley, line 32: I think you probably want an `else` to hide it otherwise, else it'll just stick with whatever the last number was, no?

christianheinrichs: @seansfkelley Thanks for your review.

AFAIK, the badge was hidden even when the `else` statement was missing. The newest commit addresses your review comment.

ke-d:
I have verified that it is fixed. Thanks for the contribution!

christianheinrichs:
You're welcome!

PR #239, opened 2017-11-13
Add link from README to documentation in the wiki

ElectricFeet:
I suggest adding link to documentation in the wiki, as this is not obvious for the average browser user who is not used to github.


kennethtran93:
I'll merge this and add it onto the current 3.X.X branch as well.

ElectricFeet:
Thanks @kennethtran93 !

PR #240, opened 2017-11-13
Add link from "About" to documentation

ElectricFeet:
A link to the documentation doesn't seem to be available at all in the extension itself.


kennethtran93:
I've added this change into the 3.X.X branch per #627 .  As you had made this for 2.X.X, I will close it as I won't be making any changes into 2.X.X.

PR #290, opened 2017-12-17
Append timestamp to backup filename.

kennethtran93:
My Implementation for feature/Issue  for appending timestamp to backup filename.

EDIT:  Latest commit makes it more seamless than my first commit.  This will append the timestamp to the filename on either left or right click.

Sample Exported Filename:
Sample Exported Filename

Some Title hints are given too upon mouse-over of 'Export Expressions' Button and the generated filename link.

Tested on Fedora 27 on Firefox 57.0.1 and Chrome 63.

Comments/Feedback/Critique/Suggestions for other improvements are welcomed.


kennethtran93:
Fixing a few commit changes.  Will issue new PR.

PR #292, opened 2017-12-18
Auto Focus on edit input

kennethtran93:
This should take care of

`autofocus` attribute is new in HTML5, so a workaround was also implemented for older HTML versions within `componentDidUpdate()`.

Since `componentDidUpdate()` runs after each component update, thus checks were added to check if we're actually editing a domain expression, and focuses on that edit input.  If the new expression input is currently in focus and an edit input is shown, it won't focus onto the edit input.

Additionally, the caret will be at the end of the current value whenever the edit input is focused.  Originally without the `onFocus` attribute and the `moveCaretToEnd` function, the caret would appear in the beginning of the edit text.  If that should be intended result, simply remove the `onFocus` attribute and `moveCaretToEnd` function.


kennethtran93:
Closing PR to resolve some commit stuff.  New branch and PR to come.

PR #294, opened 2017-12-18
Added dynamic timestamp to export filename (Resolves #216)

### LINKED ISSUE #216: Append time/date to backup filename
All backup filenames are the same.

Adding the time/date is useful for sorting and identifying specific backups. uBO does this, example format for C-AD:

`CAD_Expressions_2017-10-10_12.00.00.json` or `CAD_2017-10-10_12.00.00.json`

kennethtran93:
My Implementation for feature request #216 for appending timestamp to backup filename.

This will append the timestamp to the filename on either left or right click.

Sample Exported Filename:
Sample Exported Filename

Added some hints are given upon mouse-over of 'Export Expressions' Button.

Tested on Fedora 27 on Firefox 57.0.1 and Chrome 63.

Comments/Feedback/Critique/Suggestions for other improvements are welcomed.

### DIFF
--- a/extension/_locales/en/messages.json
+++ b/extension/_locales/en/messages.json
@@ -200,6 +200,11 @@
     "description": "Export Expressions"
   },
 
+  "exportURLSTitle": {
+    "message": "A timestamp will be appended to the filename on export.\nHaving trouble downloading?  Right-Click this button and click 'Save Link As...' to download.",
+    "description": "A timestamp will be appended to the filename on export.\nHaving trouble downloading?  Right-Click this button and click 'Save Link As...' to download."
+  },
+
   "importURLSText": {
     "message": "Import Expressions",
     "description": "Import Expressions"
--- a/src/ui/popup/App.js
+++ b/src/ui/popup/App.js
@@ -130,7 +130,7 @@ class App extends Component {
 
 						<div
 							className="btn-group"
-							ref={(e) => { this.cleanButtonContainerRef = e; }}
+							ref={(e) => {this.cleanButtonContainerRef = e;}}
 							style={{
 								margin: "0 4px"
 							}}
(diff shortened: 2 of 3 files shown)


grenzor:
Not sure if it's possible to change here but I think the filename should follow `Cookie_AutoDelete_2017-10-10_12.00.00.json` as the sample filename is too unnecessarily long.

Version number (2.X.X) and "Expressions" aren't needed in filename since all backups are the same format regardless of version and nothing but expressions can be exported anyway.

kennethtran93:
It's possible to technically name it anything - just needs general consent/approval from @mrdokenny before I make the new change and commit/push it in this branch/PR.

kennethtran93:
@mrdokenny What's the status for this PR/Feature?

ke-d:
>I'm wondering if you could use line 96's template string (or call a function that returns that string) and use it in line 153? That way you don't need a id.

kennethtran93:
I presume you're referencing the files from my commit, and I know what you mean.   However, putting the template string or calling the function within the download attribute only works ONCE per page load - whether directly or via function.  The filename timestamp won't change upon subsequent button clicks.  It needs to be called from onClick (left click) as well as onContextMenu (right click), thus a function was created to minimize the duplicate code since either left or right clicks will run the exact same code.

If you don't want an ID in the button, I may have come up with a workaround for it.  I'll commit directly to this PR in a few moments once I test and clean up the new code some more.

kennethtran93:
With the use of onClick and onContextMenu, the original download name would be overwritten anyway, so technically we don't really need it there, but I've kept it in for now.

kennethtran93:
Latest commit 4f7b63a removes the ID from button and the workaround was to have nodejs give my function the element itself.

kennethtran93:
Before this gets merged, would you tell me your answer to @grenzor's post about the shortening of the filename template?  I can easily do another commit to chance the name within this PR as well.

ke-d:
@kennethtran93 Maybe something like `CAD_Expressions_{Date here}`?

Update: I'm going to merge this anyways and change it afterwards. Thanks for the contribution.

Issue #234, opened 2017-11-13
Clicking 'Edit expression' should focus for text input


ghost:
### Describe the Bug/Feature

When I click the Edit expression button, I have to manually click the domain input box afterwards. It should focus the keyboard unto the domain input automatically

### Your settings 

- OS/version: GNU/Linux
- Browser/version: Firefox 56
- Cookie AutoDelete version: 2.0.1

ke-d:
Resolved in #295

PR #295, opened 2017-12-18
Added Focus to input field on Edit

kennethtran93:
My implementation for feature request/bugfix of #234.

`autofocus` attribute is new in HTML5, so a workaround was also implemented for older HTML versions within `componentDidUpdate()`.

Since `componentDidUpdate()` runs after each component update, thus checks were added to check if we're actually editing a domain expression, and focuses on that edit input.  If the new expression input is currently in focus and an edit input is already shown, it won't focus onto the edit input.

Additionally, the caret will be at the end of the current value whenever the edit input is focused.  Originally without the `onFocus` attribute and the `moveCaretToEnd` function, the caret would appear in the beginning of the edit text.  If that should be intended result, simply remove the `onFocus` attribute and `moveCaretToEnd` function.


ke-d:
@kennethtran93 Thanks! I'll look at this and all the other PRs soon when I have time. :)

ke-d:
Thanks for the contribution.

PR #331, opened 2018-01-21
Close popup when opening settings page

dylanhanning:
This change makes the popup automatically close when the 'settings'
button is clicked, so that the settings page is shown without the
popup partially obscuring it.

This seems a bit cleaner/nicer than needing to click around in the
settings page to get rid of the popup.

Note: I'm using the 2.X.X branch because other PRs seem to, but let me know if that's incorrect. The _contributing_ section in the README doesn't mention where to branch features from :)

### DIFF
--- a/src/ui/popup/App.js
+++ b/src/ui/popup/App.js
@@ -235,9 +235,12 @@ class App extends Component {
 							style={{
 								margin: "0 4px"
 							}}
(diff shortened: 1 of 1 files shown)


ke-d:
Thank you for the contribution!

2.X.X branch is fine for now.

Issue #311, opened 2018-01-06
Popup window size frequently becomes too small on macOS Chrome


uu1t:
### Describe the Bug

When opening popup window, frequently it appears only in an moment and then becomes too small.
This happens on Chrome for Mac, not Windows (I confirmed on Windows 10).

### Screenshot in which the bug can be seen

![popup_fails](https://user-images.githubusercontent.com/35105763/34637158-1781dcf6-f2f3-11e7-9c3d-8da66abcf6bc.gif)

### Your settings 

- OS/version: macOS v10.13.2
- Browser/version: Chrome v63.0.3239.132
- Cookie AutoDelete version:  2.0.1 and latest commit (905dc5d)

ke-d:
I don't have a Mac, so I can't verify or fix this, but I'll accept PRs for this.

uu1t:
I think this is a Chrome's issue which hasn't been fixed for several years.  
Here is the issue: https://bugs.chromium.org/p/chromium/issues/detail?id=428044

ansell:
Can confirm, happens about 1 in 5 times

ghost:
I've seen this bug occur randomly on several macOS Chrome extensions for some years, not just Cookie-AutoDelete.

ke-d:
Closing as this should have been fixed.

PR #340, opened 2018-01-28
Add delay before rendering popup on Chrome

uu1t:
Add 100ms delay before rendering popup on Chrome.
This is a workaround for #311.

### DIFF
--- a/extension/popup/popup.html
+++ b/extension/popup/popup.html
@@ -9,6 +9,7 @@
   </head>
   <body>
     <script type="text/javascript" src="../global_files/browser-polyfill.js"></script>
+    <script src="../global_files/browserDetect.js"></script>
     <script src="../global_files/react.min.js"></script>
     <script src="../global_files/react-dom.min.js"></script>
     <script src="../global_files/redux.js"></script>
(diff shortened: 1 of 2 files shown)


ke-d:
@kojole Interesting I wonder if it's caused by popup reflow? Since the `render()` method technically runs twice: 1 for the start and 2 for the async call to finish in this [line](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/blob/2.X.X-Branch/src/ui/popup/App.js#L30). This can cause a slight reflow since the site name is not populated.

Does adding this piece of code before [line 113](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/blob/2.X.X-Branch/src/ui/popup/App.js#L113) resolve your problem without your workaround?
```
		if (Object.keys(tab).length === 0) {
			return ("Loading");
		}
```

uu1t:
No. I added and tested this code, but it doesn't work well.

It seems that there is some time (about 50ms, I think) after popup's `DOMContentLoaded` when Chrome doesn't handle DOM size changes correctly. So DOM changes need to be delayed.

PR #342, opened 2018-01-30
OPERA support

rastamanf1:
_Hello,_

thank you very much for your beautiful and very useful extension. I am already using it in Chrome and Firefox, but is there a chance to have a **version for the [Opera](http://www.opera.com) browser?**

_Thanks in advance._

_All the best!_


ke-d:
Although technically it should work since it's based on similar APIs, I don't have the time to test support for three browsers.

Issue #148, opened 2017-09-13
Immidiately delete cookies when changing URL domain in open tabs


Johnny1000:
This is a request for an enhancement in minimizing tracking cookies tracking abilities.
(I hope I have understood properly how cookies can be used to track users.)

If the underlying API has support for notifications when open tabs have their URLs changed, it would be nice with an extra option to immidiately delete cookies from URL domains that are changed out of open tabs.

Currently the deletion delay can be set to 0 minutes, so cookies from a domain are deleted immidiately when the last tab containing a URL to that domain is closed.

However, if the user prerfer using the same tab for a new URL domain, and just change the URL domain in that open tab, cookies placed by the previous domain visit in that tab stay the delay time before they are deleted by the active mode. In that time tracking cookies can be read through the new URL's domain.

Immidiate deletion when changing URL domain in open tabs would avoid cookie tracking, at least as long as the user uses only one tab, but in general also every time a time overlap between tracking domains is avoided.

Currently the user can of course set the deletion delay to 0 minutes, open a new empty tab, delete the old tab with the domain whose cookies should be deleted, and then enter the new domain in the new empty tab.

But it would be a little nicer having a setting that enables automatic deletion on URL change in open tabs, so the user can just put the new URL in the same tab, and have the cookies from the old URL domain deleted immidiately when the browser starts going to the new URL domain.

Best regards
(And thank you for a good extension :o))
Johnny :o)

ke-d:
Thanks for using it. 

This request would actually take some time if I ever decide to do it since the extension would have to keep track of all open domains and compare the diff on page load. So I won't be considering this request anytime soon until I understand the performance impact of this.

Johnny1000:
Thank you for answering :o)

Doesn't the system already keep track of all open domains [1] and compare diff on tab close?
How else does the system know _not_ to delete cookies from a domain [1] still open in tabs other than the one being closed?

Try this simple test:

Test conditions:
Disable all other extensions, so they don't mess up the test.
Deletion delay set to 0 minuttes.
2 open tabs to 2 _different_ pages within the same domain [1], both pages setting cookies.

Test:
Check cookies, and note the cookies set.
Close one tab on the domain [1], leaving the other tab open on the same domain [1].
Check cookies.
-- All cookies from the first check is still there.
Close the last tab on the domain [1]
Check cookies.
-- All cookies from the domain [1] has been deleted.

[...]

PR #387, opened 2018-04-17
AutoDelete at PAGE/SITE LEAVE

rastamanf1:
_Hello, @mrdokenny,_

would it, please, be possible to make _Cookie AutoDelete_ **DELETE cookies** when **LEAVING a site ([SLD](https://en.wikipedia.org/wiki/Second-level_domain)) (within the same, still open tab)?** I think that this could solve the problem of **[TRACKING cookies](https://en.wikipedia.org/wiki/HTTP_cookie#Tracking)** (with **Automatic Cleaning** Delay set to **0 seconds**) - contextual (e. g. accommodation facility > booking system) redirecting should be considered ([NoCoin](https://github.com/keraf/NoCoin)'s temporary whitelisting might work for this purpose, since [AI](https://en.wikipedia.org/wiki/Artificial_intelligence) would probably be difficult, if not impossible, to program).

_Thanks in advance. **And thank you for your GREAT WORK!**_


ke-d:
Duplicate of #148

Issue #424, opened 2018-07-22
Error upon LocalStorage deletion if cookies for IPv6 address hosts are present


sqrt2:
### Describe the Bug/Feature

If a cookie from an IPv6 address hostname that is nominally safe to clean is in the cookie store, its `cookieProperties.hostname` property that gets added to `deletedHostnames` is empty. When `Array.from(deletedHostnames)` array is finally passed to `removeLocalStorage()`, Firefox throws an error because of the empty hostname in the array:

```Type error for parameter options (Error processing hostnames.1: Error: Invalid hostname ) for browsingData.removeLocalStorage.```

The cookie(s) causing the problem are not cleared from the cookie store either, so the problem persists until the user manually deletes the relevant cookies from the cookie store.

### Steps for anyone to reproduce the bug

A cookie from an IPv6 address host that is safe to delete needs to be in the cookie store. Then, upon closing a tab where cookies are safe to clean, no LocalStorage cleaning will take place and an error message is displayed in the console. No notifications appear either.

### Your settings 

Automatic cleaning on, cleanup on domain change on, cleanup log on, notifications on, cookie counter on, clean open tabs on startup off, container tabs support off, LocalStorage cleanup on

- OS/version: Linux
- Browser/version: Firefox 61.0.1
- Cookie AutoDelete version: 2.2.0

half-duplex:
Workaround for FF, so the addon's localstorage features are actually useful again without having to load my patched one in dev mode every browser launch:
about:preferences#privacy -> Cookies and Site Data -> Manage Data -> remove any with an IPv6 address (sorting by Site is helpful) -> Save Changes
And probably disable third party cookies and don't go directly to any IPv6 addresses that set cookies, or it'll start failing silently again...

ke-d:
Closing since 3.0.0 is out.

PR #431, opened 2018-08-08
Remove empty string from hostnamesDeleted

half-duplex:
Possibly mitigates #424
I had a problem which may not have had the same cause but produced the same error and seemed to prevent the addon from functioning at all. This handwaves the real problem away but seems to allow things to work without having to completely disable localstorage deletion for now.


half-duplex:
Obsoleted by #457

PR #457, opened 2018-10-25
Wrap v6 IPs in square brackets

### LINKED ISSUE #424: Error upon LocalStorage deletion if cookies for IPv6 address hosts are present

half-duplex:
Obsoletes #431. Should more better fix #424.

In Firefox, cookies are still not correctly cleared for IPv6 address URLs (http://[2001:db8::1234]/). This problem is not introduced by this patch, it just becomes more obvious. I believe the issue is with firefox's browser.cookies.{add,remove}() functions, and unless anyone has any ideas, I'll open a bug report with them.

Repro:
Since .add() seems broken, go to a v6 site, eg. http://[2001:19f0:5c00:8de2::64]/, and in the page console run `document.cookie="test=a";`
Then open the debug console for the addon and run `b=browser.cookies.remove({url: "http://[2001:19f0:5c00:8de2::64]/", name: "test", firstPartyDomain: ""});`
`b`, a promise, is fulfilled with value `null`, which means that no cookies were found to delete, which is incorrect. I've tried every variation I can come up with and have not been able to delete the cookie this way.


ke-d:
Looks simple enough... thanks!

PR #494, opened 2019-02-01
Version Number and Log Timezone Name Additions

kennethtran93:
Closes #443 and possibly closes #474.

(Apologies - I forgot to create a separate branch for these separate 'bugfixes'.  Hopefully you're fine with implementing both at once)

Added version number in three places:  

- Icon Title (viewable upon mouse-over or in the firefox overflow menu)
screenshot from 2019-01-31 22-02-14

- Sidebar menu navigation in Settings
screenshot from 2019-01-31 22-03-32

- About page
screenshot from 2019-01-31 22-04-30

Value is extracted from the manifest file from within the browser itself, so there's still only one version number to keep updated.

-----

For the log times, I've added in the option to display the short timezone name that is based on what is already stored by the timestamp itself.
screenshot from 2019-01-31 22-31-40

PR #503, opened 2019-02-21
Added Versioning on Notifications Popup

kennethtran93:
Thought I should do this as well, before you publish a new version.   As the title says, this adds the version number to the popup.

I am unsure if you had used the `notificationTitle` locale string anywhere else, so let me know if I need to commit additional changes to this PR first.

Removed `versionExtText` from my previous PR #494 as I found out that I used a workaround instead based on your current locales string for mentioning browser version.

For those looking for a screenshot, here you go.  Nothing too fancy.
screenshot from 2019-02-21 14-46-30

### DIFF
--- a/extension/_locales/en/messages.json
+++ b/extension/_locales/en/messages.json
@@ -10,8 +10,14 @@
   },
 
   "notificationTitle": {
-    "message": "Cookie AutoDelete: Cookies were deleted!",
-    "description": "Cookie AutoDelete: Cookies were deleted!"
+    "message": "Cookie AutoDelete $Ver$: Cookies were deleted!",
+    "description": "Cookie AutoDelete 3.X.X: Cookies were deleted!",
+    "placeholders": {
+      "ver": {
+        "content": "$1",
+        "example": "3.0.0"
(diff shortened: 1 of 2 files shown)


kennethtran93:
I'm working on propagating the locale changes to all the other languages (not sure if I needed to do that but I will anyway - please wait for another commit from me later (shouldn't be more than a day later).

ke-d:
> I'm working on propagating the locale changes to all the other languages (not sure if I needed to do that but I will anyway - please wait for another commit from me later (shouldn't be more than a day later).

You generally don't need to do that since I upload the english version to Crowdin and then it will propagate the changes.

kennethtran93:
Then let's leave it at that then.  I don't think there's any other places that uses `notificationTitle` but I could be wrong - are there any?

kennethtran93:
Any chance of getting this merged before the next version is published?

ke-d:
@kennethtran93 Sorry about this, just been pretty busy lately with other stuff. Let me know if there's anything else I'm forgetting.

PR #619, opened 2020-02-10
Fixed IP Regex Bug

### LINKED ISSUE #539: IP Regex bug (work-12345678.workplace.com) is being treated as an IP
On some websites, e.g. workplace.com (facebook's office spinoff) the presence of Cookie Autodelete addon logs out the user almoust every minute, even though it's whitelisted. When I disable the addon it works fine.

This and some others problems can be solved if there was disable on this website option, or maybe it's a whitelist bug?

- OS/version: 8.1
- Browser/version: FF 67.0.2
- Cookie AutoDelete version: 3.0.2

kennethtran93:
This fixes #539.  IP Regex Bug, at least for this specific scenario.  

As I'm using the ip-regex library, it should be much better at properly detecting valid IPs (both V4 and V6).

Added (2) additional tests for this bug.

Updated Dependencies.


kennethtran93:
Here's a quick screenshot of the patched bug

Screenshot from 2020-02-10 13-16-21

Issue #169, opened 2017-10-05
Allow users to enter their own regular expressions


grenzor:
Currently it looks like all user entered expressions are converted to regular expressions but this doesn't allow users to enter their own regular expressions. Doing so will be extremely useful for user-defined specificity for some rules.

ke-d:
If you can convince me of specific use cases, I might consider it.

As it stands, I don't see how this would be useful vs. making more confusing for the user (not everyone can do RegExp).

grenzor:
I agree this is for more expert users so users would have to do something extra to use it (e.g. tick box to enter regular expression).

^((?!google.com|github.com|yahoo.com).)*$ - matches all domains except defined

This is just one example from me, someone who really knows very little regexp but since regexp is extremely flexible I'm sure it would benefit more advanced users. Honestly, I was originally thinking of this suggestion more for your H-AD addon not C-AD as C-AD regexp is useful for cookie domains where the other can be applied to the whole URL but it can still be useful here since RegExp support is already implemented.

seansfkelley:
When I first saw the changes this was also my first thought. I don't have a use-case off the top of my head now, but I thought it was strange that the UI displayed the regex when it wasn't editable/relevant.

geniekoechling:
A use case would be a regex for the private IPv4 adresses:

`(^127\.)|(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)`

or ranges of hostnames like `srv000.example.com - srv099.example.com` without allowing every subdomain.

kennethtran93:
Those following this:  I'm working on getting the cookie list populated based on the domain regexp. It will be included in the next version.

Issue #646, opened 2020-03-08
[FEATURE REQUEST] Using C-AD as anti-distraction extension


maltsev:
I guess, the primary use-case of this extension is to prevent being tracked. But it also seems that this extension would work pretty well for a totally different issue: preventing from visiting Facebook, Twitter, Reddit, and other distraction websites too often. Since all these websites require you to be signed in to get any meaningful experience letting C-AD automatically sign you out after you close the tab would, in my opinion, significantly reduce impulse visits of this websites since each time you'll have to enter your credentials (so it isn't a quick dopamine hit anymore).

The only thing that prevents me from using C-AD for that purpose is that I don't want to be signed out of all other websites that I visit from time to time. Sure, I can add them to the white list but it'd require adding each of these websites separately, and I have probably a hundred of them.

A possible solution would be to implement some sort of black-list, so C-AD deletes only cookies from that list and keeps other cookies present (discussed in #134 and #549). 

Another solution is to add regexp support so it'd be possible to white list all websites except Facebook, Twitter, etc (discussed in #169). 

I'm a developer myself so I can implement that. But first I'd like to have a discussion if you see this use-case fit for C-AD at all and if you eventually ready to accept my PR.

kennethtran93:
A blacklist on top of the existing whitelist/greylist is fine.  It would be implemented as always delete then.   Another option on the settings will need to be added such that one can set a delay on the blacklist itself.  This may be related to #631 .

Regex support could be possible too.  I was going to implement the ability to use comma-separate site value sometime soon.

With an additional button in the popup, we will need to make it a dropdown selection, maybe even need to adjust the list of expressions page too.

maltsev:
Thanks for the clarification! I'll check this weekend which one is better to implement.

kennethtran93:
Is this a closable feature request or will additional items be implemented?

maltsev:
The solution implemented in #649 works well for me, but maybe it will be too complex for other people to use. On the other hand, I don't plan to work on this issue anymore, so I guess, we could close it for now. 

I'll send you updated docs about this feature this evening.

kennethtran93:
Thanks!  Version 3.1.0 is uploaded for review to both extension stores, but if you want to try you can download it from GitHub Releases.

maltsev:
> I'll send you updated docs about this feature this evening.

It seems you already took care of that :-)

PR #649, opened 2020-03-15
Add regexp support for white-list rules

maltsev:
In #646 we discussed how C-AD could be used as an anti-distraction extension. I decided to choose the easiest path and just add regexp support to white-list rules. In the UI it looks the following way: 
image

By default, it's unchecked so users who don't know what regexps are could just simply ignore that.

I've tested this change and it works well both in FF and Chrome.

I was not sure if it's a good place in the UI to put that option. Maybe you could suggest a better place?

I could also update the wiki if this PR gets merged.

_(this PR also relates to #169)_


kennethtran93:
Thanks!

Would it be more feasible if it could just check if the expression entered is a regex? (i.e. check for `/` as the first character and `/` as the end character? That would eliminate the RegExp checkbox.

kennethtran93:
Heads up #650 is my PR with a bunch of changes.  As it is late night for me I won't merge my PR yet just in case I missed something in the morning.

maltsev:
Thanks for letting me know! Can you wait a bit with releasing a new version of C-AD? I'll implement the regexp the way you suggested in ≈10 hours.

maltsev:
I've added check for `/` as a first and a last char.

kennethtran93:
While the regexp works in what you have given me, I'm also fixing a bug that surfaced due to using regexp as the 'keep all cookies' unchecked would show nothing.  This is because the cookies.getAll was expecting an actual domain and not a regexp.  I'm going to add an additional check and filter to the 'keep all cookies' unchecked so it would only show cookies if the domain matches the regexp given.

kennethtran93 (approved):
Thanks for the addition of testing with the enhancement.

kennethtran93:
Will Close #169 with this

kennethtran93:
Thanks!

PR #682, opened 2020-04-11
Adding support for the great suspender extension

intika:
Adding support for the great suspender extension (klbibkeccnjlkjkiokjodocebajanakg)

When automatic feature is enabled cookie are deleted for suspended tab as well which should not be the case, this PR fix it.

### DIFF
--- a/src/services/Libs.ts
+++ b/src/services/Libs.ts
@@ -19,13 +19,24 @@ export const getHostname = (urlToGetHostName: string | undefined) => {
   if (!urlToGetHostName) {
     return '';
   }
-  try {
-    // Strip "www." if the URL starts with it.
-    const hostname = new URL(urlToGetHostName).hostname.replace(
+  try { 
+ 	// Adding support for the great suspender extension (klbibkeccnjlkjkiokjodocebajanakg)
+	 if (new URL(urlToGetHostName).hostname == 'klbibkeccnjlkjkiokjodocebajanakg') {
+	   const hostnameSuspenderExt = urlToGetHostName.replace(
+      /^chrome-extension:\/\/klbibkeccnjlkjkiokjodocebajanakg.*uri=/,
+      '',
+      );
+		  return new URL(hostnameSuspenderExt).hostname.replace(
       /^www[a-z0-9]?\./,
       '',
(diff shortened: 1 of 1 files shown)


intika:
Closing this PR... feel free to reopen/use/merge it if the fix suit your requirement.

kennethtran93:
Thanks for your contribution.  It would certainly be a first for this extension to include checks for third-party extensions.  

However there are other ways of keeping cookies by using grey list for the ones that you want to keep until at least the browser restarts, or white list if you want to keep them forever.    Since the tab would have been an active one then you could try to make it part of your checks to either grey list or white list the domain from the popup.  

To undo the clearing of cookies, you must have already enabled the 'Enable Cleanup Log and Counter' option.  This would only affect future cleanups.

To make it a more global item, perhaps an option could be implemented such that if it detects that the tab is an internal / webextension URL, then check the entire URL for an actual http/https domain and go from there.

Thoughts?

intika:
> To make it a more global item, perhaps an option could be implemented such that if it detects that the tab is an internal / webextension URL, then check the entire URL for an actual http/https domain and go from there.

Yes this would probably make it compatible with more extension then just "the great suspender" but extracting (guessing) the domain url with regex may probably not work all the time as each extension do the work differently.

This feature may come handy if it's implemented with an option on the settings, but i think that most great applications/extensions does keep the settings to minimum for efficiency and simplicity. 

When the "auto clean" feature is enabled the rule is simple, "keep cookies only for opened tabs" (i guess so), so if i have a suspended tab that is still open i would keep its cookie... (suspending tabs is just useful to speedup the browser, when a multitude of tabs are opened)... Gray list feature in this case require additional manipulation for suspended tabs and thus we kind a loose the "auto" thing...

I like your extension a lot as it's the only one that is truly working (chrome)... i use to have my own extension for the matter especially for Firefox... i switched to yours because i have a limited time for this task, and if i can contribute, why not ;)

Anyway, bottom line, the outcome here (at least for me) was to have it working with my config... i guess that an advanced/good implementation would be the same with just a different regex to extract any domain from any extension like tab (chrome-extension://)

PR #721, opened 2020-05-10
Clean option for discarded tabs

Grumph:
Not working yet, as I don't understand how the cookies are actually deleted.

Fix #720

PR #753, opened 2020-06-01
Right click menus PLUS openTabs by container and discarded tab cleanup

### LINKED ISSUE #262: Cookie protection should not work across containers
Closing a tab in one container does not delete that tab's cookies if the same domain is still open in a different container. That doesn't make sense to me. Since the purpose of containers is to isolate users' activities, CAD should respect that isolation. When deciding whether to delete a closed tab's cookies, CAD should only check for open tabs in the same container.

### Steps to reproduce

1. In your Firefox preferences, make sure you have at least one container set up besides the default container. If you're creating a new container, initialize your CAD settings for it as described [here](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/issues/197#issuecomment-337523756).
2. Choose a domain (`example.com`) that normally stores cookies and is not whitelisted in your CAD prefs. Enable automatic cleaning with no delay in your CAD prefs.
3. Make sure you don't have any tabs open to `example.com`.
4. Make sure your Firefox prefs are set to allow cookies from `example.com`, and if you have any `example.com` cookies stored already, delete them.
5. Open a new tab (Tab A) in any container, including the default container, and go to `example.com`.
6. Check to make sure that `example.com` has actually stored cookies, and then clear them, either with the CAD "Clear cookies for this domain" button or manually in your Firefox prefs.
7. Open a new tab (Tab B) in any other container and go to `example.com`.
8. Check to make sure that `example.com` has stored cookies again, and then close Tab B.

Expected results: CAD instantly deletes `example.com` cookies from Container B's cookie store.
Actual results: CAD deletes `example.com` cookies from Container B's store only after Tab A is closed.

### Your settings 

- OS/version: macOS 10.12.6
- Browser/version: FF 57.0
- Cookie AutoDelete version: 2.0.1
All settings checked except Show Notification After Cookie Cleanup?

### LINKED ISSUE #386: Always show notification when manually clearing cookies
I think it would be useful to show the notification when manually clearing cookies via the toolbar icon, even if the notification is disabled in the settings.

This way the user has some certainty that the button he just clicked had an effect.

Thanks!

### LINKED ISSUE #415: Feature Request: One Click Cookie Delete Toolbar Shortcut
I'm finding it a hassle to delete cookies using this add on. I would like the ability to add a shortcut to the toolbar that allows me to press "Clean, include open tabs" in just one click. Normally that would require three steps instead of just one;

1. Press the Icon
2. Open the Clean drop down menu
3. Press "Clean, include open tabs"

Can you guys just simply allow us to add a toolbar button that works in a single click? 
Toolbar Buttons would include;
"Clean, include open tabs"
"Clear cookies for this domain"
"Clear localstorage for this domain"
"Clean"

Thanks.

EDIT: I may suggest that if you right click the proposed shortcut icon, it brings down the normal Cookie AutoDelete Menu.

### LINKED ISSUE #521: Add option to whitelist/graylist domain from logs
It would be really great if I am able to add a domain to whitelist or gray list directly from the logs. This is particularly useful when using extensions such as evernote or lastpass.

### LINKED ISSUE #716: [BUG] [Firefox Containers] "Automatic cleaning" triggers only when site's tabs are closed in all containers
**Describe the bug**

The "Enable automatic cleaning with delay of [x] seconds" main feature activates when all tabs using a site are closed.

Unfortunately, this literally means "all tabs", not "all tabs within this container".

This results in cookies for sites that are visited in more than one container not being cleaned properly.

**Steps to reproduce**

CAD Settings:
* Enable Automatic Cleaning? Yes
* Delay Before Cleaning? 15 seconds (default)

Firefox containers:
* Default container / no container
* * CAD Expressions: none, clean everything
* Container "Test"
* * CAD Expressions: whitelist for *.github.com

Test 1:

1. Open tab in container "Test". Log in to Github in this tab.
2. Close all tabs of container "Test".
3. Wait >15 seconds, refresh CAD log. There should be no cleanup entry for github.
4. Open new tab in container "Test", navigate to Github, confirm you are still logged in.

**OK**

Test 2:

1. Close all tabs of container "Test".
2. Open tab in default container. Log in to Github in this tab.
3. Close all default continer tabs.
4. Wait >15 seconds, refresh CAD log. Cleanup entry for github in default container should appear.
5. Open new tab in default container, navigate to Github, confirm you are NOT logged in.

**OK**

Test 3:

1. Log in to github again in the default container. Leave the tab open for now.
2. Open new tab in container "Test". Navigate to Github. Leave the tab open.
(You should still be logged in due to whitelist; that's probably not relevant to this bug but let's make note of that).
3. Close all github tabs in default container, leave github tabs in container "Test" open.
4. Wait >15 seconds, refresh CAD log. There should be no cleanup entry for github.
5. Open new tab in default container, navigate to Github, confirm you are still logged in.
6. Close all github tabs, both in default container and in "Test" container.
7. Wait >15 seconds, refresh CAD log. Cleanup entry for github in default container should appear.
8. Open new tab in default container, navigate to Github, confirm you are NOT logged in.

**BUG**

**Expected behavior**
When all tabs for a site are closed in Container 1, its cookies should be cleaned inside Container 1 following the auto-clean timer and whitelist rules set up.
Existence of open tabs for the same site in container 2, 3, or 75 - other identities - should be irrelevant.
After all the white/grey list is per-container.

**Observed behavior**
The auto-clean task kicks off only when all tabs for a site are closed, globally in the browser, across all containers.
The cookies for the site are then cleaned or preserved in accordance to each container's whitelist.

**System Info**

 - Windows 10 Pro 1909
 - Firefox 76.0
 - Multi-Account-Containers 6.2.5
 - Cookie AutoDelete 3.2.0

### LINKED ISSUE #717: Right Click support
**NEED:** I am frustrated when I sign-on to certain browser extensions (e.g., Evernote that provide a check box to stay logged in for 30-days that opens a window) but for which, inside the sign-in pop-up window, I cannot whitelist the URL that is shown in the popup.  

I have tried going to the apparent URL that it looks like it is going to inside the pop-up window and white listed that page, but then when I go to use the app again after shutting down the browser, it has lost its cookie so presume it is not the correct page.

**WISH:** What I would like to be able to do is Right Click mouse and see a place to whitelist the URL with Cookie Autodelete, as this would allow me to whitelist insight these types of popup windows.

     I cannot envision an alternative solution, but am open to anything that would relieve the overhead of having to sign-on so often for apps I use all day everyday.  Thanks for listening.

### LINKED ISSUE #720: [FEATURE REQUEST]  clean cookies for discarded tabs
**Is your feature request related to a problem? Please describe.**

I use extensions that discard/unload tabs from memory (AutoTabDiscard to name one).
When a tab is discarded, it's unloaded from memory, and reloaded later.
I expect that cookies are removed for discarded tabs, as in my workflow it's the same if I clos the tab and reopen it later.

**Describe the solution you'd like**

Launch the clean timer when the tab is discarded.

**Describe alternatives you've considered**

I tried to build a solution by myself, but it's the first time I read TypeScript and do WebExtension developpement, and I don't understand how the cookies are deleted.

**Additional context**

See attached draft. #721

kennethtran93:
- Closes #717.
- Closes #521 in regards to adding expression via right-click-selection-text.
- Closes #415 in regards to right-click shortcuts.
- Closes #386 (Notifications on manual clean of cookies/localstorage from popup or right-click).
- Opening settings will now open next to the current tab.
- Reworked CleanupService tests.  The previous decreasing coverage was getting to me.

Updated 2020 June 05
- Closes #720 (Tab Discard), based on initial draft PR #721.
- Closes #716 and Closes #262 (Open Tabs per container).


kennethtran93:
Testers needed - right-click menu tests.

[Cookie-AutoDelete_Dev_20200601_181157_Chrome.zip](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/files/4712997/Cookie-AutoDelete_Dev_20200601_181157_Chrome.zip)
[Cookie-AutoDelete_Dev_20200601_181157_Firefox.zip](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/files/4712998/Cookie-AutoDelete_Dev_20200601_181157_Firefox.zip)

kennethtran93:
@ke-d Can you give this a test run and see if it all works for you?  If you have any suggestions for improvement I'd like to hear it.

kennethtran93:
Realistically this build shouldn't have any drastic changes aside from the unit tests...

[Cookie-AutoDelete_Dev_20200605_032814_Chrome.zip](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/files/4733741/Cookie-AutoDelete_Dev_20200605_032814_Chrome.zip)

[Cookie-AutoDelete_Dev_20200605_032814_Firefox.zip](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/files/4733748/Cookie-AutoDelete_Dev_20200605_032814_Firefox.zip)

Feedback/Suggestions from testing is appreciated...before this gets mass-released.

kennethtran93:
As it was somewhat easier to continue enhancements after refactoring the CleanupService tests, the completed enhancements has been included in this PR.

- Option to clean tabs that have tab.discarded set true.
- Bugfix: Open Tabs are now grouped by containers.

[Cookie-AutoDelete_Dev_20200605_194111_Chrome.zip](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/files/4738228/Cookie-AutoDelete_Dev_20200605_194111_Chrome.zip)
[Cookie-AutoDelete_Dev_20200605_194111_Firefox.zip](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/files/4738229/Cookie-AutoDelete_Dev_20200605_194111_Firefox.zip)

__tests__/services/CleanupService.spec.ts
```diff
@@ -104,116 +150,644 @@ const mockCookie: CookiePropertiesCleanup = {
...
+        const ffResult = await cleanCookiesOperation(firefoxState);
+        expect(spyLib.isFirstPartyIsolate).toHaveBeenCalledTimes(1);
+        expect(global.browser.contextualIdentities.query).not.toHaveBeenCalled();
+        expect(global.browser.extension.isAllowedIncognitoAccess).toHaveBeenCalledTimes(1);
+        expect(global.browser.cookies.getAllCookieStores).toHaveBeenCalledTimes(1);
+        // getAll called from: isFirstPartyIsolate x1, directly with storeId: 'default' and 'firefox-default';
+        expect(global.browser.cookies.getAll).toHaveBeenCalledTimes(3);
+        expect(global.browser.cookies.remove).toHaveBeenCalledTimes(2);
+        expect(global.browser.browsingData.removeLocalStorage).not.toHaveBeenCalled();
+
```
ke-d: To be honest, if some or all of the `toHaveBeenCalled();` lines can be removed without decreasing coverage, then I rather not have it to not clutter the tests.

The only time it makes sense is that if you are testing something important like `        expect(global.browser.cookies.getAll).toHaveBeenCalledTimes(3);` where the input would affect this in the CleanupService.

__tests__/services/CleanupService.spec.ts
```diff
@@ -104,116 +150,644 @@ const mockCookie: CookiePropertiesCleanup = {
...
+      it('Regular clean, exclude open tabs (Firefox) to include errors during cleanCookies / browser.cookies.remove (via cadLog)', async () => {
+        when(global.browser.cookies.remove)
+          .calledWith(expect.any(Object))
+          .mockRejectedValue(new Error('test') as never);
+        await cleanCookiesOperation(firefoxState, cleanupProperties);
+        // isFirstPartyIsolate plus one for each storeId: 'default', 'firefox-default'
+        expect(global.browser.cookies.getAll).toHaveBeenCalledTimes(3);
+        // this error always call cadLog even if debugMode is disabled.
+        expect(spyLib.cadLog).toHaveBeenCalledTimes(1);
```
ke-d: I'm not sure if it's necessary to spy the cadlog in here and other places.

__tests__/setup.js
```diff
@@ -0,0 +1,150 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+'use strict';
+
```
ke-d, line 26: Can you add a comment explaining the purpose of this file?

src/background.ts
```diff
@@ -151,14 +166,330 @@ const onStartUp = async () => {
...
+  }
 };
 
+async function onContextMenuClicked(
+  info: browser.contextMenus.OnClickData,
+  tab: browser.tabs.Tab
+) {
+  const debug = getSetting(store.getState(), 'debugMode') as boolean;
+  const contextualIdentities = getSetting(store.getState(), 'contextualIdentities') as boolean;
+  if (debug) {
```
ke-d: Instead of if (debug) statements everywhere, might I suggest just calling 

```
    cadLog({
      msg: `background.onContextMenuClicked:  Data received`,
      x: {info, tab},
    });
```
And then inside the cadLog function, it would have the `if (debug)` statement. Doing this should reduce the amount of test cases as well since there's no more condition.

src/background.ts
```diff
@@ -151,14 +166,330 @@ const onStartUp = async () => {
...
+  browser.browserAction.setTitle({ title: `${mf.name} ${mf.version} [READY] (0)` });
+  if (currentSettings.debugMode.value) {
+    cadLog({
+      msg: `background.onStartUp is complete.`,
+      type: "info"
+    });
+  }
 };
 
+async function onContextMenuClicked(
```
ke-d: There's a lot of code here, can we move this function to another file?

src/redux/Actions.ts
```diff
@@ -271,12 +271,12 @@ export const cookieCleanup: ActionCreator<
...
   // Show notifications after cleanup
   if (
-    setOfDeletedDomainCookies.size > 0 &&
+    setOfDeletedDomainCookies.length > 0 &&
     getSetting(getState(), 'showNotificationAfterCleanup')
   ) {
     const notifyMessage = browser.i18n.getMessage('notificationContent', [
       recentlyCleaned.toString(),
-      Array.from(setOfDeletedDomainCookies).join(', '),
+      setOfDeletedDomainCookies.join(', '),
```
ke-d, line 279: Why is this needed?

ke-d (changes requested):
Thanks for the PR. Would recommend 1 PR per 1 issue resolved next time. :)

src/background.ts
kennethtran93: So what you're saying is to pass another variable into cadLog, call it debug, and then in each of the calling functions pass in ```getSetting(store.getState(), 'debugMode') as boolean```?

src/background.ts
kennethtran93: I'll move it into ContextMenuEvents.

src/redux/Actions.ts
kennethtran93: This was because when I was reading the output the cadLog gave, it only returned `{}` when passed in as a Set. Well, `returnSetOfOpenTabDomain` now returns an object with the container id as the key, then the open tab domains.    As such I modified it in `returnSetOfOpenTabDomains` to return an array instead of a Set.  

It doesn't necessarily need to be changed, but if an array is already being returned, there's no point of converting an array into an array again.

__tests__/services/CleanupService.spec.ts
kennethtran93: With coverage, at least one test is needed to cover the functions called, including cadLog, depending on how the path to it is given.

__tests__/services/CleanupService.spec.ts
kennethtran93: I'll try and minimize the amount of tests, but coverage requires all functions to be covered, irregardless of where it affects the extension...

src/background.ts
kennethtran93: Heads up - when I'm done working on this, more files will be changed.

Normally I'd say if a function doesn't need to be called then use `if`, but if you want to minimize test cases then we'll do it that way then.

__tests__/services/CleanupService.spec.ts
kennethtran93: As the previous test have already covered it, I'll remove redundancy ones.

src/background.ts
ke-d: > So what you're saying is to pass another variable into cadLog, call it debug, and then in each of the calling functions pass in getSetting(store.getState(), 'debugMode') as boolean?

Yeah that would work.

Eventually though we may want to do some dependency injection (i.e. make cadLog into a singleton class function that has the state of the extension) so we don't have to pass in  `getSetting(store.getState(), 'debugMode')`. But your solution will work for the short term.

kennethtran93: I'd haven't thought about moving cadLog from Libs to a separate class...I had thought to keep the multiple use functions into Libs.

Mind you this was only an initial debug log - I had intend for it to either show up on a new settings page for easier debugging/exporting.

Also in the future this debug should have levels as there is so much information being sent through.

kennethtran93: The only other thing is that there are some cadLog calls that still outputs regardless of debug mode, so if we're moving it into a separate class that needs to be taken into consideration.

src/redux/Actions.ts
kennethtran93: Marking as resolved per above comment, unless changes in `returnSetOfOpenTabDomains` should be renamed now that it isn't returning a set...

__tests__/services/CleanupService.spec.ts
kennethtran93: Should be resolved through 2b5e3b2

__tests__/setup.js
kennethtran93: Should be resolved through 2b5e3b2

src/background.ts
kennethtran93: Should be resolved through 2b5e3b2, at least through adding in the parameter for now.

src/background.ts
kennethtran93: Should be resolved through 2b5e3b2

src/redux/Actions.ts
kennethtran93: In 342cc64 I've renamed `returnSetOfOpenTabDomains` to `returnContainersOfOpenTabDomains`  
The function now returns an object, with the container id as key and an array of open tab domains.

src/services/Libs.ts
```diff
@@ -238,6 +257,13 @@ export const localFileToRegex = (hostname: string) => {
   return hostname; // Doesn't have a file path...return as is.
 };
 
+/**
+ * Parse cookieStoreId for use in addExpressionUI...
+ */
+export const parseCookieStoreId = (contextualIdentities: boolean, cookieStoreId: string | undefined) => {
```
ke-d, line 263: Minor nit: I think this can be `cookieStoreId: string?`

src/services/CleanupService.ts
```diff
@@ -222,79 +205,154 @@ export const cleanCookies = async (
...
+  if (cookies.length > 0) {
+    let cookieDeletedCount = 0;
+    for (const cookie of cookies) {
+      const r = await browser.cookies.remove(returnOptionalCookieAPIAttributes(state, {
+        firstPartyDomain: cookie.firstPartyDomain,
+        name: cookie.name,
+        storeId: cookie.storeId,
+        url: prepareCookieDomain(cookie),
+      }, firstPartyIsolate) as {
+        // Fix type error with undefineds with cookies.remove
```
ke-d, line 232: Interesting, I wonder why is this?

src/services/CleanupService.ts
```diff
@@ -222,79 +205,154 @@ export const cleanCookies = async (
...
+        url: string;
+        name: string;
+      });
+      if (r !== null) cookieDeletedCount += 1;
```
ke-d: `if (!r)` ?

ke-d (dismissed):
Some minor stuff and I don't see any other major problems so approved.

src/services/Libs.ts
kennethtran93: While I'd like to, that is actually an optional parameter per https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/Tab, not to mention Chrome doesn't have one.  The check within the function checks if it is defined otherwise assigns 'default' when returning for CAD lists.

Also when I tried removing it just now TypeScript is complaining, even though we'll always have the appropriate firefox permission to always have cookieStoreId as a parameter.

src/services/CleanupService.ts
kennethtran93: What are you referring to here?

src/services/CleanupService.ts
kennethtran93: yeah I can change that.  I was just following MDN where it says it will return null if it wasn't removed.

kennethtran93: so I guess it should be `if (r)` instead.

kennethtran93: resolved via 6a1a285

src/services/CleanupService.ts
kennethtran93: I'll keep this PR/branch until we clarify what is needed here.

ke-d: Why does `browser.cookies.remove` return undefined here? Is the return type not defined in the types library?

kennethtran93: I'm not sure - I had extracted the code that used to be in the popup, so whatever you had done at that time.

Upon further research, the `url` isn't defined in cookies as the url that is required in `cookies.remove` consists of the combination of domain + path that we take from cookies.  Because url and name is required, it doesn't take in `undefined` within the type, unlike our other custom types that were extended from `browser.cookies.Cookie` with optional parameters

kennethtran93: also because `returnOptionalCookieAPIAttributes` has an input type of `Partial<CookiePropertiesCleanup>` which I believe applies `| undefined` to all parameters.

wrkq:
Hi. Sorry for late comment.
The cleanup of cookies per container seems to work exactly as it should now, after testing the basic couple of scenarios. It does not seem to accidentally mess up cookies in whitelisted containers or anything like that either.
Thank you a ton for this. It's really nice to not have to manually clean up when that one dumb social media link or something opens in current container.

PR #763, opened 2020-06-10
Dependency/TypeScript Types Updates

kennethtran93:
- @types/jest from 25.2.3 to 26.0.0.
- @types/react from 16.9.35 to 16.9.36.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2193,9 +2193,9 @@
       }
     },
     "@types/jest": {
-      "version": "25.2.3",
-      "resolved": "https://registry.npmjs.org/@types/jest/-/jest-25.2.3.tgz",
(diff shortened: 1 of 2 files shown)


kennethtran93:
@ke-d Normally I'd just directly push it through the branch but since you're back I may as well do it properly.

ke-d (approved):
Hmm maybe we should set up a periodic (monthly?) PR where we update dependencies and do some manual testing instead of doing minor updates whenever there's a new update, since you never know what could break.

Of course this policy would be overridden by security patches and such.

kennethtran93:
Sure.

In this case, it is just typescript items, and all tests appear to be fine.  I've attempted to look over the commits made to @types/jest and appears to have just add in an overloaded function that jest added in 26.

PR #765, opened 2020-06-12
Unit Tests for TabEvents.

kennethtran93:
- Enhancements to TabEvents to make it slightly easier for coverage.
- Minor enhancement to AlarmEvents so that we can use jest.spyOn.
- Minor enhancement to BrowserActionService to include a check and to use proper camelCase in a function.
- Removed a semi-colon in an interface in Cleanup.d.ts.
- Removed unnecessary continue from CleanupService.spec.ts
- Fix typo in setup.js and simplified functions.

### DIFF
--- a/__tests__/services/TabEvents.spec.ts
+++ b/__tests__/services/TabEvents.spec.ts
@@ -20,6 +20,7 @@ import { initialState } from '../../src/redux/State';
 import createStore from '../../src/redux/Store';
 import { ReduxAction, ReduxConstants } from '../../src/typings/ReduxConstants';
 import AlarmEvents from '../../src/services/AlarmEvents';
+import * as BrowserActionService from '../../src/services/BrowserActionService';
 import * as Lib from '../../src/services/Libs';
 import TabEvents from '../../src/services/TabEvents';
 import StoreUser from '../../src/services/StoreUser';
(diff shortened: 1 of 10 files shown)


__tests__/services/TabEvents.spec.ts
```diff
@@ -0,0 +1,394 @@
...
+    // Do not change any of the test order as each test relies on the previous actions.
+
+    it('should do nothing if tab.status is not complete', () => {
+      TabEvents.onDomainChange(0, sampleChangeInfo, {...sampleTab, status: 'loading'});
+      expect(spyTabEvents.cleanFromTabEvents).not.toHaveBeenCalled();
+    });
+
+    it('should set mainDomain on first encounter', () => {
+      TabEvents.onDomainChange(0, sampleChangeInfo, {...sampleTab, status: 'complete'});
+      expect(spyLib.cadLog.mock.calls[0][0].msg.indexOf('First')).not.toBe(-1);
```
ke-d: Like before, I'm not sure if we need spy libs for cadLogs and test for whether or not it was called, since it would mean the TabEvents class would depend on on cadLog.

Wouldn't it be the same coverage since cadLog would be unit tested independently?

Same comments whenever spylib for cadLog gets called.

kennethtran93: For onDomainChange at least should be needed as there is no other way of expecting something as the tabToDomain variable is set to private and jest can't check private variables unless I were to make another temporary class that extends TabEvents just to check/expose that private variable...

While I can just call the functions for coverage, there's not really a point unless something is expected from it.

kennethtran93: Also I'm not just checking if cadLog was called but the parameters passed into it.

kennethtran93: I'll try the temporary class method just like what I did for `StoreUser` in the morning.

kennethtran93: Mostly resolved through 6f01422.  

There are still (3) calls to spyLib.cadLog due to the fact that I'm sanitizing a variable that is useless in debug output.

ke-d (approved):

kennethtran93:
Thanks!

PR #769, opened 2020-06-14
Migrate ts-lint to es-lint due to deprecation

ke-d:
Looks like the test finally pass after messing around with the config files. 🤷 

Anyway [TSLint](https://palantir.github.io/tslint/) is deprecated as of 2019, this PR migrates over to eslint.

Most of the file changes were prettier file formats and fixes for the new *simplified* .eslintrc.json.

I fixed all of the errors but left the warnings in the default config for now since we can fix them at anytime.

Most of these changes should be Typescript related so there shouldn't be any user facing changes or bugs. Still good to check manually in beta phase for a couple of days as a sanity check. 😄


ke-d:
Looks like coverage is failing but not too concerned about that right now, since the tests still pass.

__tests__/services/CleanupService.spec.ts
```diff
@@ -68,10 +68,10 @@ const wildCardGreyGit: Expression = {
   id: '4',
   listType: ListType.GREY,
   storeId: 'default',
-}
+};
 
 const whiteListAllExceptTwitter: Expression = {
-  expression: '/^((?!twitter\.com).)+$/',
+  expression: '/^((?!twitter.com).)+$/',
```
kennethtran93: This is a regex expression - the backslash needs to stay.

__tests__/services/CleanupService.spec.ts
```diff
@@ -304,39 +310,61 @@ describe('CleanupService', () => {
...
+        );
+        expect(
+          global.browser.contextualIdentities.query,
+        ).not.toHaveBeenCalled();
+        expect(
+          global.browser.extension.isAllowedIncognitoAccess,
+        ).toHaveBeenCalledTimes(1);
+        expect(global.browser.cookies.getAllCookieStores).toHaveBeenCalledTimes(
+          1,
+        );
```
kennethtran93, line 348: Interesting how eslint/prettier made this into separate line(s).

__tests__/services/TabEvents.spec.ts
```diff
@@ -28,19 +28,21 @@ import StoreUser from '../../src/services/StoreUser';
 jest.requireActual('../../src/services/AlarmEvents');
 
 // This dynamically generates the spies for all functions in AlarmEvents
-const spyAlarmEvents: {[s: string]: jest.SpyInstance} = {};
+const spyAlarmEvents: { [s: string]: jest.SpyInstance } = {};
 for (const k of Object.keys(AlarmEvents)) {
```
kennethtran93: As said previously, last PR moved the spy generator into setup.js.  With the formatting difference, you'll need to adjust it yourself.

src/background.ts
```diff
@@ -167,34 +178,43 @@ const onStartUp = async () => {
...
-    if (!browser.contextMenus.onClicked.hasListener(ContextMenuEvents.onContextMenuClicked)) {
-      browser.contextMenus.onClicked.addListener(ContextMenuEvents.onContextMenuClicked);
+    if (
+      !browser.contextMenus.onClicked.hasListener(
+        ContextMenuEvents.onContextMenuClicked,
+      )
+    ) {
+      browser.contextMenus.onClicked.addListener(
+        ContextMenuEvents.onContextMenuClicked,
+      );
```
kennethtran93: Not sure if we can have prettier config change so it doesn't quite look like this?  That or I just need to try and get used to it somehow.

tsconfig.json
```diff
@@ -1,40 +1,62 @@
 {
   "compilerOptions": {
     /* Basic Options */
-    "target": "es6",
-    "module": "commonjs",
+    "target": "es6" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
+    "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
     "lib": [
       "es2017",
```
kennethtran93, line 7: Your es file is using es2020, can anything be adjusted here?

src/ui/settings/components/SettingsTooltip.tsx
```diff
@@ -19,9 +19,11 @@ interface OwnProps {
 }
 
 const SettingsTooltip: React.FunctionComponent<OwnProps> = ({ hrefURL }) => {
-  const link = (hrefURL.startsWith('http') ? hrefURL : `https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/wiki/Documentation${hrefURL}`);
+  const link = hrefURL.startsWith('http')
+    ? hrefURL
+    : `https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/wiki/Documentation${hrefURL}`;
   return (
-    <a href={link} target={'_blank'} className="tooltipCustom">
+    <a href={link} target="_blank'" rel="noreferrer" className="tooltipCustom">
```
kennethtran93:
```suggestion
    <a href={link} target="_blank" rel="noreferrer" className="tooltipCustom">
```
Removing the lone single quote.  should be `_blank` for target I believe

src/ui/common_components/IconButton.tsx
```diff
@@ -34,11 +34,19 @@ interface IconButtonProps {
...
+      tag,
+      text,
+      ...nativeProps
+    } = this.props;
 
     // Has to be PascalCase, else JSX will think it's a tag named 'tagName'.
     const TagName = tag === 'input' ? 'label' : tag || 'button';
     return (
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore
```
kennethtran93, line 50: What's happening with these ignore lines?  Can we just have the eslint line itself or do we really need ts-ignore?

src/ui/common_components/ExpressionOptions.tsx
```diff
@@ -40,7 +43,7 @@ const styles = {
 };
 
 const trimDotAndStar = (str: string) => {
-  const trimmed = str.replace(/^[\.\*]+|[\.\*]+$/g, '');
+  const trimmed = str.replace(/^[.*]+|[.*]+$/g, '');
```
kennethtran93, line 46: Regular Expression.  Please keep the backslash escapes.

src/services/Libs.ts
```diff
@@ -403,8 +428,7 @@ export const throwErrorNotification = (e: Error) => {
 /**
  * Trim leading and ending dot of a string
  */
-export const trimDot = (str: string) => str.replace(/^[\.]+|[\.]+$/g, '');
-
+export const trimDot = (str: string) => str.replace(/^[.]+|[.]+$/g, '');
```
kennethtran93, line 431: Regular Expression.  Keep Backslashes for escapes.  Otherwise `.` means any one character.

src/services/Libs.ts
```diff
@@ -77,15 +78,15 @@ export const createPartialTabInfo = (
...
+  };
+};
 
 /**
  * Converts a version string to a number
  */
 export const convertVersionToNumber = (version?: string) => {
   if (!version) return -1;
-  return parseInt(version.replace(/[\.]/g, ''), 10);
+  return parseInt(version.replace(/[.]/g, ''), 10);
```
kennethtran93, line 89: Regular Expression.

src/redux/Reducers.ts
```diff
@@ -16,16 +16,15 @@ import shortid from 'shortid';
...
 const hasExpression = (
   list: ReadonlyArray<Expression>,
   action: { payload: Expression },
-) => list.some(expObj => expObj.expression === action.payload.expression);
+) => list.some((expObj) => expObj.expression === action.payload.expression);
 
 // Creates a new Expression object to be stored in the list
 const newExpressionObject = (
-  state: Expression | {},
+  state: Expression | Record<string, unknown>,
```
kennethtran93, line 27: Record is something new to me...I guess any object is technically a Record...as long as we know the type?

.eslintrc.json
```diff
@@ -0,0 +1,37 @@
...
+      // default to "createReactClass"
+      "pragma": "React", // Pragma to use, default to "React"
+      "version": "detect" // React version. "detect" automatically picks the version you have installed.
+    }
+  },
+  "plugins": ["react", "@typescript-eslint"],
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "off",
+    "react/prop-types": ["off"],
+    "no-console": "warn"
```
kennethtran93, line 35: I'll trust you on these rules as long as most of them were ported over from tslint.

kennethtran93 (commented):
Some are comments, others are questions or requesting changes, notably the missing backslash for escaping characters in Regular Expressions.

__tests__/services/CleanupService.spec.ts
ke-d: https://eslint.org/docs/rules/no-useless-escape

> Escaping non-special characters in strings, template literals, and regular expressions doesn't have any effect

The tests seem to pass just fine, so I think we're ok here

tsconfig.json
ke-d: So all I did for this file was just reverted to the tsconfig I had long ago and that worked lol.

I don't plan on making any future tweaks to this file in this PR as least.

__tests__/services/CleanupService.spec.ts
kennethtran93: I think the check may be treating the expression as a string instead of its intended regex...
I suppose.
Isn't a dot a special character?  It stands for any one character.
  It may just look weird because if we're manually reading this the regex would mean that 'twitterxcom' would match...

__tests__/services/CleanupService.spec.ts
ke-d: Ack.

src/background.ts
ke-d: Eh I trust its defaults.

src/services/Libs.ts
ke-d: See Above.

src/services/Libs.ts
ke-d: See above.

src/redux/Reducers.ts
ke-d: Eslint recommended this change.

.eslintrc.json
ke-d: Well I didn't want to mess around too much with the new defaults. So the first two,

```
"@typescript-eslint/no-explicit-any": "off",
"react/prop-types": ["off"],
```
will eventually be remove once we refactor some more, but this is a good start for now.

src/ui/common_components/IconButton.tsx
ke-d: ts-ignore is for the TS compiler, while the eslint is for ignoring the ts-ignore for eslint to function.

src/ui/common_components/ExpressionOptions.tsx
kennethtran93: resolved as eslint fix good.

__tests__/services/CleanupService.spec.ts
ke-d: https://stackoverflow.com/questions/47277133/disable-unnecessary-escape-character-no-useless-escape

For the most part it seems safe to remove the `\`, but if we need to, we can use `[.]` instead.

kennethtran93: Because we accept regex and string in the expression, eslint doesn't know that.  I think this line should be the exception to the no-useless-escape eslint.

kennethtran93: > https://stackoverflow.com/questions/47277133/disable-unnecessary-escape-character-no-useless-escape
> 
> For the most part it seems safe to remove the `\`, but if we need to, we can use `[.]` instead.

I would agree with this had the \ be within the `[]`, but there wasn't any.

kennethtran93: I guess for testing purposes and visual look the [.] should be done then.  Seems a bit excessive for an extra character compared to `\.` but if that will make eslint not warn, then sure.

kennethtran93 (approved):
Let's do it.  I'll get another alpha/beta pre-release build out once this and my other PRs are merged.

Go ahead and override the failing coverage this time.

...or should I do that..?

PR #774, opened 2020-06-15
Hopefully fixes Icon not staying blue if no cookies are found.

kennethtran93:
...If you're wondering, both `cookieLength = undefined` and `cookieLength = 0` both return `false` in conditionals.


kennethtran93:
[Cookie-AutoDelete_Dev_20200615_221457_Chrome.zip](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/files/4783259/Cookie-AutoDelete_Dev_20200615_221457_Chrome.zip)
[Cookie-AutoDelete_Dev_20200615_221457_Firefox.zip](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/files/4783260/Cookie-AutoDelete_Dev_20200615_221457_Firefox.zip)

kennethtran93:
LOL the coverage increased because I fixed a bug...

src/services/BrowserActionService.ts
```diff
@@ -175,7 +175,7 @@ export const checkIfProtected = async (
           break;
       }
     } else {
-      if (cookieLength && cookieLength === 0) {
+      if (cookieLength !== undefined && cookieLength === 0) {
```
ke-d, line 178: I don't understand. Isn't doing a truthy check or cookieLength good enough? As it is checking for both undefined and null?

kennethtran93: I had left a note in the PR comment but I'll re-iterate here:

when cookieLength is undefined OR defined as number 0, both return false in conditionals.

ke-d (approved):

src/services/BrowserActionService.ts
kennethtran93: oh wait...did I just correct myself there...

PR #782, opened 2020-06-24
Added option to purge Expired Cookies.

### LINKED ISSUE #344: Expired/Session Cookies Not Cleared
### Expired/Session Cookies Not Cleared

Date expired and session cookies are not cleared if the tab containing them is open at startup, either by setting the option to remember open windows and tabs or by pinning the tab.

Cookies for sites not open at startup are handled as expected.

The "Clean Cookies from Open Tabs on Startup" cleans up cookies from other associated sub-domains but does not solve the problem.

I first noticed this through sites like Twitch which can set CSRF session tokens for individual paths on the site and Youtube witch can set session cookies with the ID of the video if you have multiple pages open at once. The number of cookies stored for both sites was increasing over time for me and further investigation showed the old session cookies and cookies past their expiration date.

I have verified this behavior on a newly created profile with only Cookie AutoDelete installed.

### Steps for anyone to reproduce the bug
- Create a clean profile
- Install Cookie AutoDelete
- Visit a site that sets session cookies and whitelist the site.
- Examine a session cookie for the site to verify it exists.
- Restart the browser and verify the session cookie has been deleted but others remain.
- Set the browser option to remember open windows and tabs or open the test site and pin the tab.
- Examine a session cookie for the site to verify it exists.
- Restart the browser and verify the session cookie has not been deleted and contains the same data from the previous session.

### Versions
- OS/version: Windows 10 1709
- Browser/version: Firefox 58.0.1
- Cookie AutoDelete version: 2.1.0

kennethtran93:
Closes #344
- Default is false to keep current behavior.
- **This ignores all open tabs and list configurations.**
- Tried to use formatting specified via Prettier.
- Also in here:  bugfix for restoring cookies when debug mode is enabled.
  - Need to deep copy objects otherwise it just references the object...to which we don't want as we want to sanitize the cookie value for debug.

### DIFF
--- a/src/services/CleanupService.ts
+++ b/src/services/CleanupService.ts
@@ -40,12 +40,18 @@ export const prepareCookie = (cookie: browser.cookies.Cookie, debug = false) => 
     cookieProperties.hostname = cookieProperties.preparedCookieDomain;
     cookieProperties.mainDomain = cookieProperties.preparedCookieDomain;
   } else {
-    cookieProperties.hostname =  getHostname(cookieProperties.preparedCookieDomain);
+    cookieProperties.hostname = getHostname(cookieProperties.preparedCookieDomain);
     cookieProperties.mainDomain = extractMainDomain(cookieProperties.hostname);
   }
   cadLog({
     msg: 'CleanupService.prepareCookie: results',
-    x: { domain: cookie.domain, path: cookie.path, preparedCookieDomain: cookieProperties.preparedCookieDomain, mainDomain: cookieProperties.mainDomain, hostname: cookieProperties.hostname},
+    x: {
+      domain: cookie.domain,
+      path: cookie.path,
+      preparedCookieDomain: cookieProperties.preparedCookieDomain,
+      mainDomain: cookieProperties.mainDomain,
+      hostname: cookieProperties.hostname,
+    },
   }, debug);
   return cookieProperties;
 };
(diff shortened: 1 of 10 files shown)


src/services/CleanupService.ts
```diff
@@ -363,15 +497,35 @@ export const cleanLocalstorage = (bool?: boolean) => {
...
   );
   const nonBlankCookieHostName = obj.cookie.hostname.trim() !== '';
-  cadLog({
-    msg: 'CleanupService.filterLocalstorage: debug data.',
-    x: {notProtectedByOpenTab, notInAnyLists, listCleanLocalstorage, nonBlankCookieHostName, clean1: notInAnyLists, clean2: (notProtectedByOpenTab && listCleanLocalstorage) , CleanReasonObject: obj},
-  }, debug);
+  let cro: CleanReasonObject = obj;
+  if (debug) {
+    // We need deep copying object to as to not change actual cookies
+    const sanitized: CleanReasonObject = JSON.parse(JSON.stringify(obj));
```
ke-d: Can't you use object spread here?

`const sanitized: CleanReasonObject = {...obj};`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

src/services/CleanupService.ts
```diff
@@ -481,62 +640,95 @@ export const cleanCookiesOperation = async (
...
+        prepareCookie(cookie, debug),
+        newCleanupProperties,
+      );
     });
 
     if (debug) {
-      const sanitized = isSafeToCleanObjects.map(obj => {
-        obj.cookie.value = "***";
+      // We need deep copying object to as to not change actual cookies
+      const debugObj: CleanReasonObject[] = JSON.parse(
```
ke-d: Same thing as before.

src/services/CleanupService.ts
```diff
@@ -481,62 +640,95 @@ export const cleanCookiesOperation = async (
...
+        debug,
+      );
       return r;
     });
 
     if (debug) {
-      const sanitized = markedForDeletion.map(obj => {
-        obj.cookie.value = "***";
+      // We need deep copying object to as to not change actual cookies
+      const debugObj: CleanReasonObject[] = JSON.parse(
```
ke-d: Same reason as before.

ke-d (changes requested):

src/services/CleanupService.ts
kennethtran93: That itself won't do it due it being only shallow copied, which means that the inner object would only be a reference to the actual object, thus would restore cookies with value of `'***'`.

I went one step further and found that this works:
```js
    const sanitized: CleanReasonObject = {
      ...obj,
      cookie: { ...obj.cookie, value: '***' },
    };
```
so I'll adjust accordingly.

kennethtran93: Should be resolved through 4cb01b4

src/services/CleanupService.ts
kennethtran93: Resolved through 4cb01b4

src/services/CleanupService.ts
kennethtran93: Resolved through 4cb01b4

kennethtran93:
@ke-d also fixed some ESLint warnings that came through on those files, namely functions with no explicit return type.

ke-d (approved):

PR #784, opened 2020-06-24
Tooltip link rel + setting popup containers

kennethtran93:
- Add `help noopener` to `<a rel...> `
- Have Firefox Desktop open settings tab in the same container as where the popup was triggered.

### DIFF
--- a/src/ui/popup/App.tsx
+++ b/src/ui/popup/App.tsx
@@ -28,6 +28,8 @@ import {
   getHostname,
   getSetting,
   isAnIP,
+  isChrome,
+  isFirefoxNotAndroid,
   isFirstPartyIsolate,
   localFileToRegex,
   LSCLEANUPNAME,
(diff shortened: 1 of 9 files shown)


src/ui/popup/App.tsx
```diff
@@ -362,10 +362,23 @@ class App extends React.Component<PopupAppComponentProps, InitialState> {
             iconName="cog"
             className="btn-info"
             onClick={() => {
-              browser.tabs.create({
-                index: tab.index + 1,
-                url: '/settings/settings.html#tabSettings',
-              });
+              if (
+                cache.browserDetect &&
```
ke-d: Optional: I feel like a lib function `isFirefox` and `isChrome` where you pass in the cache and return true or false would be useful here, since this browser detection would be used a lot.

ke-d (dismissed):

src/ui/popup/App.tsx
kennethtran93: Done via c66c6ea

kennethtran93:
...somehow the additional tests that I did got removed in the later commits O.o

kennethtran93:
Need to merge 782 first.

ke-d (dismissed):

kennethtran93:
Previously approved before rebasing from 3.X, so merging as is.

PR #793, opened 2020-07-02
Fixed Clean dropdown border

Nautman:
# Problem

The button doesn't look very neat.

Rounded corners on clean buttons dropdown

# Fix

I fixed it so that in the popup, the clean dropdown menu button has no rounded corners.

It now looks like this:
Not rounded corners on clean buttons dropdown

Compare this to the [dropdowns on Bootstraps official documentation](https://getbootstrap.com/docs/4.0/components/dropdowns/):

Dropdown on bootstraps documentation

# Potential new fix / commit
There is one benefit with the unrounded corners, and that is that it's very obvious that the menu is a separate thing from the actual button. On GitHub, they solve this by adding a border:
GitHub dropdown button

If you would like, I could add such a border that would look like this:

Border dividing dropdown menu

To add this, I would simply add a color to the left border.

```
borderLeftColor: '#c69500',
```

### DIFF
--- a/src/ui/popup/App.tsx
+++ b/src/ui/popup/App.tsx
@@ -300,6 +300,8 @@ class App extends React.Component<PopupAppComponentProps, InitialState> {
                 data-disabled="true"
                 role="menu"
                 style={{
+                  borderBottomLeftRadius: 0,
+                  borderTopLeftRadius: 0,
                   transform: 'translate3d(-3px, 0px, 0px)',
                 }}
               >


Nautman:
In hindsight I probably should have made this a draft pull request, since I'm wondering if you think it should have a colored left border... Then it also wouldn't have to perform all the checks again.

kennethtran93:
I think the original idea came from this: https://getbootstrap.com/docs/4.1/components/dropdowns/#split-button but wasn't implemented properly.

kennethtran93:
I'm going to take a look at reorganizing the html to make it work, but thanks for the initial contribution.

src/ui/popup/App.tsx
```diff
@@ -300,6 +300,8 @@ class App extends React.Component<PopupAppComponentProps, InitialState> {
                 data-disabled="true"
                 role="menu"
                 style={{
+                  borderBottomLeftRadius: 0,
+                  borderTopLeftRadius: 0,
```
kennethtran93, line 304: Found a better way - just remove the div "dropdown" parent, and add 
``` 
borderLeftColor: 'rgb(176, 132, 0)',
``` 
to the style (removing the two radius lines you added) and maybe change '-3px' to '-2px' for somewhat uniform spacing in translate3d.

See if that works better for you.

kennethtran93 (changes requested):
Try out the commented instructions and see if it works better.  I think on my end it seems to work fine.

You can have the credits for this once you made the adjustments.

kennethtran93:
superseded with PR #798

PR #797, opened 2020-07-05
Fix help link for Keep Default Icon on all list types

szpak:
The new one seems to work:
https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/wiki/Documentation#keep-default-icon-on-all-list-types


kennethtran93:
Whoops thanks for catching that.

kennethtran93 (approved):

Issue #362, opened 2018-02-24
Container-Specific Expressions are not cleaned up


stoically:
Hi, thanks for your work on Cookie AutoDelete!

### Describe the Bug/Feature

When disabling all container-relevant Add-ons, the containers are reset. After that, the internal `cookieStoreId` is reused, again starting with `firefox-container-1`. So if containers are reset, Expressions defined by Cookie AutoDelete can end up in the wrong container

### Steps for anyone to reproduce the bug

1. Fresh profile
2. Install latest CAD
3. Create Container in `about:preferences#containers` named "Test"
4. Open Tab in "Test" container and navigate to example.com
5. Set Expression for example.com
6. Disable CAD
7. Create Container in `about:preferences#containers` named "Test2"
8. Open Tab in "Test2" container and navigate to example.com
9. Check the expressions

![cad](https://user-images.githubusercontent.com/29637501/36631562-60c34e2c-1972-11e8-86d5-76d1e8d14ba9.gif)

### Your settings 

Container support enabled

- OS/version: Ubuntu 16.04
- Browser/version: Firefox 58.0.2
- Cookie AutoDelete version: 2.1.0

### Note

[`contextualIdenties.onRemoved`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/contextualIdentities/onRemoved) could help with that. If you identify this as a bug I could look into providing a PR

Related: https://github.com/mozilla/multi-account-containers/issues/864

Best regards

ke-d:
contextualIdentities events are only supported in FF 57 so I would have to bump it the min version for next version.

Also, is the cookieStoreId being reused in Firefox or in CAD?

stoically:
The cookieStoreId is reused from Firefox if the containers are reset, so they start with `firefox-container-1` again, or more specifically with `firefox-container-5`. However, I just realized that `onRemoved` wouldn't help in the case described in this Issue since that event isn't fired when containers are reset, and even if it got fired, CAD would be disabled in this case and couldn't react to it.

If containers are removed normally (without reset), the cookieStoreId isn't reused, so that's no problem.

So this is actually an edge-case and I don't see any sane way for CAD to work around that. I think the way to go would probably be to file an Firefox platform bug to request that container ids are unique through the lifetime of a Firefox profile.

cecoates:
Containerise is interesting in that it seems to look for the container names, and it creates them if they don't exist yet.

[...]

PR #798, opened 2020-07-06
Additional Site Data Type Cleanups.

### LINKED ISSUE #171: IndexedDB support
Aside from #44, persistent information can also be stored in IndexedDB and should be treated equally.
Is it possible to do this in FF57?

### LINKED ISSUE #264: Flash Cookies or LSO
I was wondering if this addon supports deleting and preventing the deletion of flash cookies or LSO's. If it doesn't could you consider extending it's functionality please?

### LINKED ISSUE #275: Little Problems and Suggestions with the Interface/Documentation/Help
Hey, I use this Add On, since my old Add on "Self-Destructing Cookies"is not available anymore in the new Firefox (Quantum).
But I had some Problems understanding the Program and the Interface, so here is my 5 Cents about this:

- In the README.md it says `Enable “Active Mode” in the popup or settings`. I only find "Auto Clean Enabled/Disabled". I assume this just needs to be updated.

- The Popup of the `Auto Clean Enabled/Disabled` Button is "Toggle to Disable/Enable automatic Cleanup". This does not help at all. I'd rather suggest something like "Automatic Cleanup deletes all Cookies after the Browser is closed"
The same goes for the other Buttons: 
`Clean`: "Run Cleanup Now" But it should be "Deletes all unwanted Cookies NOW, excluding openTabs"
`Clean, include open Tabs` Same as above.

- There is a nice FAQ here on Github: [Documentation/FAQ](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/wiki/FAQ:-Common-Questions-and-Issues), but I can't find the link in the Pop-Up and in the settings it is hidden in `Cookie Settings` -> *Question Mark Button*. I'd put the on top of the `Welcome` page of the settings.

- The FAQ (see above) is already quite good, but I think for a Newbie, there should be a short explanation of the terms "Clean" and "Greylist/Whitelist". Maybe even a short paragraph about what a "Cookie" is. At the moment some of this information is on the README.md, but this is hard to find. In overall i think the Documentation is a bit to focused on experienced Web users.

Best Regards and thanks for developing Open Source Software!

- OS/version: Linux 
- Browser/version: FF 57 64Bit
- Cookie AutoDelete version: I don't know where to find this information

### LINKED ISSUE #283: Feature request: delete ETag cookies
Love this addon, it would be perfect if it also blocked ETag "cookies". More info & demo at http://lucb1e.com/rp/cookielesscookies/

### LINKED ISSUE #287: Delete service workers
### Describe the Bug/Feature

Maybe it would also be useful to clear the service workers too. They can cache things or so, but likely also be used to store "tracking cookie"-like things.

And they can run in the background when websites are not open. One can open `about:serviceworkers` (or `chrome://inspect/#service-workers` in Chrome/ium) to look what service workers you already run.

### LINKED ISSUE #339: Notify that localStorage was cleaned for a website
Currently, the notification showed is worded as something like "3 cookies deleted" so it's unclear whether localStorage was was on or deleted from the website. Is it possible to add in the notification an indication of localStorage (like "3 cookies & localStorage deleted for ...") whenever that is also deleted?

### LINKED ISSUE #551: Option to clean cache
I just read this http://lucb1e.com/rp/cookielesscookies/ and thought it would be nice if Cookie AutoDelete would delete the cache for the site as well as cookies and localstorage, so this tracking method would stop to work.

### LINKED ISSUE #694: [FEATURE REQUEST]  persistent storage - delete when 0 cookies
**Is your feature request related to a problem? Please describe.**

Firefox seems to store Persistant Storage data even when cookies are deleted by CAD.

PS is obviously used for tracking. If a site cannot set, or has cookies destroyed, it will simply store them (or even more data) in PS.

This is obviously very bad.

**Describe the solution you'd like**

When the CAD script runs, kill off any remaining PS containers.

**Describe alternatives you've considered**

None. Ive only just noticed this behaviour. Ive not seen it mentioned in any other addons, nor a quick web search seems to raise this as an "issue", even though its an obvious vector for tracking.

**Additional context**

Nil. Just enjoying CAD and wish to enhance it without feature creep.

### LINKED ISSUE #779: [BUG]  Notifications showing from manual cleaning, despite being disabled.
This just began today. Haven't seen these notifications for years. 

Notifications are shown as being disabled when I click on the extension button in my app window (big red "notifications disabled") and under the extension settings ("show notification after cookie cleanup" is unticked). 

However, as of today, if I click the extension button in my app window and then click either "clear cookies" or "clear storage", I get a MacOS notification from this extension reporting the outcome. 

Please make it stop, thank you. 

* Cookie AutoDelete version 3.4.0
* Brave version 1.10.93
* MacOS version 10.15.5

### LINKED ISSUE #792: FireFox 78 allows WebExtensions to Clear by Domain: IndexedDB, ServiceWorkers, cache and pluginData
Firefox 77 supported clearing IndexDB by hostname if that helps: https://bugzilla.mozilla.org/show_bug.cgi?id=1551301.

Firefox 77 also supported clearing ServiceWorkers by hostname: https://bugzilla.mozilla.org/show_bug.cgi?id=1632990.

After that, Firefox 78, coming to an update today or tomorrow, will support clearing cache and pluginData by hostname: https://bugzilla.mozilla.org/show_bug.cgi?id=1636784.

Will consider looking at your addon once it implements these feautures

kennethtran93:
Closes #792 as parent Issue to additional site data types cleanable by hostnames.

This overhauls the localStorage auto cleanup function and generalizes it for additional site data types:
- Cache (Closes #551 and Possibly Closes #283)
  - Firefox 78+ & Chrome.
- IndexedDB (Closes #171)
  - Firefox 77+ & Chrome.
- Plugin Data (Closes #792, Closes #264 re: Flash Cookies)
  - Firefox 78+ & Chrome.
- Service Workers (Closes #287)
  - Firefox 77+ & Chrome.

*For backwards compatibility, existing expressions will default to keep these new site data types*

Additional items:
- Updated Settings section relating to site data types.
- Returns domains cleaned per site data type to be added to activity log (FUTURE COMMITS/VERSIONS/UPDATE).
- Tests included.
- Fixed some eslint warnings.
- Made Expression Options show 'until restart' localized when listType is GREY.  Should fix some initial user understandings.
- Clean Button dropdown styling update. (alternate result from #793)

New Changes committed as of July 16th:
- New Default Expression Options are now moved to an actual expression entry itself.  See related documentation.
- (Automatic) Site Data Cleanup triggers are now reflected in Activity Log Entries.
- Notifications for Site Data Cleanup - manual and automatic.  This resolves Issue/Request #339.
- Undo button will not show for entries with no cookies to restore.
- Remove individual log entries.
- Migration of settings between 3.4.0 and 3.5.0 - should work if for any reason a downgrade is required, minus the added items.

New Changes committed on July 20th:
- Documentation and FAQ links added to Welcome Page. Resolves #275

New Changes committed on July 23rd:
- Fixed notification regression bug since last commit.
- Added option for showing manual notifications (clean site data for a domain).  Closes #779.
- Reworded some English strings to include other site data other than Cookies as we will have evolved.


kennethtran93:
For those who want visual changes:

- Updated Settings
Screenshot_20200707_121650

- Updated Expression Options.
Screenshot_20200706_164544

- Updated Clean Button Styling
Screenshot_20200706_164709

- Updated Popup / Alternate Clean / Manual Cleaning
Screenshot_20200707_121826

- Updated Context Menu
Screenshot_20200707_121752

NOTE:  The `Clean All for This Domain` only cleans the listed site data we (will) support at this time due to the availability of cleaning by hostname, not anything else.

src/services/CleanupService.ts
```diff
@@ -404,110 +406,218 @@ export const clearLocalstorageForThisDomain = async (
   }
 };
 
-/** This will use the browsingData's hostname/origin attribute to delete any extra browsing data */
-export const otherBrowsingDataCleanup = async (
+export const clearSiteDataForThisDomain = async (
   state: State,
-  domains: string[],
+  siteData: string,
```
ke-d: Maybe an enum here would be better?

src/services/CleanupService.ts
```diff
@@ -404,110 +406,218 @@ export const clearLocalstorageForThisDomain = async (
...
+      domains,
+      debug,
+    );
+  }
+  return true;
+};
+
+export const removeSiteData = async (
+  siteData: string,
+  isChrome: boolean,
```
ke-d: Hmm, maybe not making it specific here would be best (not boolean). Something like a browser enum would be more futureproof for future browsers. Because otherwise, your options here are Chrome and Not Chrome.

Then in the body you can do a switch statement with the different implementations.

src/services/CleanupService.ts
```diff
@@ -404,110 +406,218 @@ export const clearLocalstorageForThisDomain = async (
...
+  ) {
+    browsingDataResult.cache = await cleanSiteData(
+      'Cache',
+      isSafeToCleanObjects,
+      chrome,
+      debug,
+    );
+  }
+  if (
+    getSetting(state, 'indexedDBCleanup') &&
```
ke-d, line 568: Don't have to do it in this PR but just looking at this, I think `indexedDBCleanup` and others for getSetting should be part of an enum string. Otherwise, it would be prone to misspellings and refactoring nightmares (like changing the name).

Just something to note for the future.

src/services/CleanupService.ts
```diff
@@ -754,42 +874,44 @@ export const cleanCookiesOperation = async (
...
-        true,
-      );
-      // if it reaches this point then cookies were deleted, so don't return undefined
-      throwErrorNotification(e);
+    // Handle all other browsingData cleanups.
+    const storeResults = await otherBrowsingDataCleanup(
+      state,
+      isSafeToCleanObjects,
+    );
+    if (storeResults.cache) {
```
ke-d: Probably be best to make a helper function here so you don't have to declare `setOfDeletedDomainX` above and repeat the logic here many times.

Like the input would be `storeResults.cache` =>`setOfDeletedDomainCaches`

Another thing is that you can do `new Set(array)`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set

src/services/Libs.ts
```diff
@@ -344,12 +344,15 @@ export const parseCookieStoreId = (
 /**
  * Prepare Domains for all cleanups.
  */
-export const prepareCleanupDomains = (domain: string): string[] => {
+export const prepareCleanupDomains = (
+  domain: string,
+  isChrome: boolean,
```
ke-d: Same situation as before. Browser enum and switch would be more extendable.

src/services/ContextMenuEvents.ts
```diff
@@ -447,6 +493,78 @@ export default class ContextMenuEvents extends StoreUser {
...
+        return;
+      }
+      cadLog(
+        {
+          msg: `ContextMenuEvents.onContextMenuClicked triggered Clean Site Data (${siteData}) For This Domain.`,
+        },
+        debug,
+      );
+      switch (siteData) {
+        case 'All':
```
ke-d, line 486: String Enum suggestion here.

src/services/CleanupService.ts
```diff
@@ -404,110 +406,218 @@ export const clearLocalstorageForThisDomain = async (
...
+ * Filters incoming objects with the site data to clean.
+ * @param siteData The site data type
+ * @param cleanReasonObjects Objects returned from isSafeToClean()
+ * @param isChrome True if Chrome Browser
+ * @param debug True if debug mode.
+ */
+export const cleanSiteData = async (
+  siteData: string,
+  cleanReasonObjects: CleanReasonObject[],
+  isChrome: boolean,
```
ke-d: Ditto as before.

ke-d (changes requested):

src/services/CleanupService.ts
kennethtran93: So what you're thinking is to create an array of objects returned from `otherBrowsingDataCleanup`, then outside the container loops, parse and combine the domains from each site data type from all containers and use Set once to remove dups?

src/services/CleanupService.ts
kennethtran93: Implemented in commit ac90490

src/services/CleanupService.ts
kennethtran93: Implemented in commit ac90490

src/services/CleanupService.ts
kennethtran93: Implemented in commit ac90490

src/services/ContextMenuEvents.ts
kennethtran93: Implemented in commit ac90490

src/services/CleanupService.ts
kennethtran93: Implemented in commit ac90490

src/services/Libs.ts
kennethtran93: Implemented in commit ac90490

kennethtran93:
New Changes committed as of July 16th:

-  New Default Expression Options are now moved to an actual expression entry itself. See related documentation.
-  (Automatic) Site Data Cleanup triggers are now reflected in Activity Log Entries.
-  Notifications for Site Data Cleanup - manual and automatic.    This resolves Issue/Request #339.
-  Undo button will not show for entries with no cookies to restore.
-  Remove individual log entries.
-  Migration of settings between 3.4.0 and 3.5.0 - should work if for any reason a downgrade is required, minus the added items.


Screenshot_20200716_160513
Screenshot_20200716_160546
Screenshot_20200715_223345

Screenshot_20200716_160733

kennethtran93:
Commit eb77f38 and 994ce6e adds the following Expression List Enhancements  (only if Firefox Contextual Identities/Containers enabled):

- Ability to remove all expressions from a single selected list.
- Now shows lists whose id could not be found in the Firefox container list.

This should help for #362

Screenshot_20200716_195344

src/services/CleanupService.ts
```diff
@@ -443,15 +467,25 @@ export const clearSiteDataForThisDomain = async (
...
+  state: State,
+  siteData: SiteDataType,
+  browserDetect: string,
   domains: string[],
   debug: boolean,
 ): Promise<boolean> => {
-  const listName = isChrome ? 'origins' : 'hostnames';
+  const listName = ((b: string) => {
+    switch (b) {
+      case 'Chrome':
```
ke-d: Optional for this PR. This can be an enum.

sdht0:
Thanks a lot for these changes!

I tested the new cleanup feature on one website and I'm not sure if CAD is not working or it's still a limitation of the API. Steps:
- Set "Enable Automatic Cleaning" and all "Other Browsing Data Cleanup Options" to true in CAD settings.
- Go to Firefox preferences -> Privacy & Security -> Manage Data... -> Search for "fandom" -> Remove selected -> Save changes.
- Refresh and open "Manage Data..." again to verify that no data for "fandom" is present.
- Open [https://onepiece.fandom.com/wiki/Toki_Toki_no_Mi](https://onepiece.fandom.com/wiki/Toki_Toki_no_Mi). This sets cookies and other storage.
- Refresh and open "Manage Data..." to verify that cookie and storage data is now present for "fandom".
- Close the website. CAD will automatically delete all data.
- Refresh and open "Manage Data...". It can be seen that all cookies were deleted successfully for "fandom", but some storage is still left behind. I am not sure how to check what kind of storage this is.

Some other feedback and suggestions:
* I'm seeing a lot of `Some site data, including 0 Cookie(s), removed from: (Private)` in the cleanup log with no detailed content. Maybe some more context/details can be added as to what was cleaned?
* In "List of Expressions" settings page, instead of putting the container id in all the tabs, maybe display it above the expression table only for the currently selected tab? I feel the tabs are now quite cluttered, esp when there are a lot of containers. Maybe also put "Remove All Expressions from Selected List" in the previous line?
* Minor: instead of "attempted to cleanup...", maybe say "invoked cleanup of..." or something equivalent that is more definitive?

src/services/CleanupService.ts
```diff
@@ -443,15 +467,25 @@ export const clearSiteDataForThisDomain = async (
...
+        return 'origins';
+      case 'Firefox':
+      default:
+        return 'hostnames';
+    }
+  })(browserDetect);
+  const sd = `${siteData[0].toLowerCase()}${siteData.slice(1)}`;
```
ke-d: I think this is common enough to be its own mini function, for the purposes of documentation and easier maintenance in the future.

src/services/Libs.ts
```diff
@@ -176,6 +184,41 @@ export const getHostname = (urlToGetHostName: string | undefined): string => {
...
+ * @param state The State (store.getState())
+ * @param storeId The container id, or 'default'
+ * @param listType The List Type
+ */
+export const getContainerExpressionDefault = (
+  state: State,
+  storeId: string,
+  listType: ListType,
+): Expression => {
+  const getExpression = (list: string): Expression | undefined => {
```
ke-d, line 200: I haven't done TS in a while, but I think that you can do `Expression?` rather than `Expression | undefined`. If this is wrong, feel free to ignore this comment.

ke-d (dismissed):

kennethtran93:
@sdht0 Thank you very much for assisting with the test builds!  I've left some notes below for you:

>   * Refresh and open "Manage Data...". It can be seen that all cookies were deleted successfully for "fandom", but some storage is still left behind. I am not sure how to check what kind of storage this is.

At this point, the only way I know of at this point, is to have that tab open, use the `Clean, include open tabs` action (via popup clean dropdown or right-click),  First, press `F12` or goto to either the `Tools` menu or the firefox hamburger menu near the top right, then click on `Web Developer` followed by `Storage Inspector`.  Here you should see the list of Cache, Cookies, LocalStorage, IndexedDB, and Session Storage.  Service Workers are under the 'Applications' tab of the same devtools panel.


> Some other feedback and suggestions:
> 
>   * I'm seeing a lot of `Some site data, including 0 Cookie(s), removed from: (Private)` in the cleanup log with no detailed content. Maybe some more context/details can be added as to what was cleaned?

If you're using Private Browsing, that is most likely the cause.  For privacy reasons, we do not output any urls that came from a Private Browsing container.  Though if that is showing up then I think I need to adjust the urls from the site data domains to not include private browsing results.

> 
>   * In "List of Expressions" settings page, instead of putting the container id in all the tabs, maybe display it above the expression table only for the currently selected tab? I feel the tabs are now quite cluttered, esp when there are a lot of containers. 

Unfortunately, due to the fact that we can store outdated container id that doesn't have an actual corresponding Firefox container (thus no name), it won't be easy.  This PR contains the recently committed feature of showing the unresolved container IDs for end-users to remove themselves, especially when exporting and importing from different profiles and/or resetting the container preferences in Firefox.

> Maybe also put "Remove All Expressions from Selected List" in the previous line?

This button only appears in Firefox when containers/contextual Identities is enabled, which falls in line with the container tabs shown as well.  There is already another button to remove all expressions which is used for all lists (if containers is enabled) for the default list if containers disabled or in Chrome.

>   * Minor: instead of "attempted to cleanup...", maybe say "invoked cleanup of..." or something equivalent that is more definitive?

Sounds like a plan.

src/services/CleanupService.ts
kennethtran93: Sure.  As long as we don't modify the return string from browserDetect.js in global_files

src/services/Libs.ts
kennethtran93: What you're referencing is a function return type, so that doesn't work apparently.

sdht0:
> Here you should see the list

It looks like both Session storage and Cache storage are not being cleaned up. Not sure about the former but at least the later should've been cleaned, no?

> If you're using Private Browsing, that is most likely the cause.

I am not using it at all. In fact, I've CAD disabled in private browsing mode as it will all get cleaned up anyways. So these logs are more mysterious in my case.

> Unfortunately, due to the fact that we can store outdated container id that doesn't have an actual corresponding Firefox container (thus no name), it won't be easy.

How about using the names when available and using the id when missing? My goal was to reduce the display clutter somehow, but am not too attached to the solutions I proposed.

> This button only appears in Firefox when containers/contextual Identities is enabled

Yeah, I was suggesting another cosmetic change to reduce the use of vertical space. Moving the button to the previous line will help avoid the white space waste to the right of the buttons. The button logically belongs to the previous line. Of course, in  other browsers it won't be displayed at all.

kennethtran93:
> It looks like both Session storage and Cache storage are not being cleaned up. Not sure about the former but at least the later should've been cleaned, no?

From the domains shown in the Cache section, what remains after cleanup?  The actual domain or third-party domains?  At this point because we don't view network requests, we don't (or rather can't) set any cleanup cookies for third-party domains.  This is already a known issue since somebody else reported that third-party localstorage doesn't get cleared if no cookies from third-party is set.  As for the former, I'm not sure either as there's no specific option for that (only localStorage).

> I am not using it at all. In fact, I've CAD disabled in private browsing mode as it will all get cleaned up anyways. So these logs are more mysterious in my case.
If you expand the summary, do you see any specific URLs from any specific site data that was cleaned up?

> How about using the names when available and using the id when missing? My goal was to reduce the display clutter somehow, but am not too attached to the solutions I proposed.

While this may work, people can set duplicate names thus only the id is unique.  I'll try and move the ID somewhere else, but it will still be shown regardless.  This is to help users that want to move expressions as until I implement that feature, users need to manually export, change the file itself, then re-import it.  The Import only relies on the container ID.

> Yeah, I was suggesting another cosmetic change to reduce the use of vertical space. Moving the button to the previous line will help avoid the white space waste to the right of the buttons. The button logically belongs to the previous line. Of course, in other browsers it won't be displayed at all.

I'll give that a shot.

sdht0:
> The actual domain or third-party domains

The actual domain. It should be pretty easy to reproduce at your end as per the steps I outlined above and help you debug if the issue is on CAD's end.

>  If you expand the summary, do you see any specific URLs from any specific site data that was cleaned up?

The details are empty (the first one is expanded):
image

>  move the ID somewhere else, but it will still be shown regardless

Yes this is what I meant. Instead of showing it for all the tabs at once, only show it for the active tab (just above the expression table is one option). Again, this was just a suggestion and I can live with the current implementation if it will mean shipping these changes faster.

kennethtran93:
> > The actual domain or third-party domains
> 
> The actual domain. It should be pretty easy to reproduce at your end as per the steps I outlined above and help you debug if the issue is on CAD's end.

Unfortunately there's not much we can debug about these.  With the exception of cookies, no other site data can be queried/enumerated through the webextension API.  All we do is pass in the hostname to be cleaned, and the type of site data to be cleaned.  (API in question is this:  https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browsingData/remove)

> 
> > If you expand the summary, do you see any specific URLs from any specific site data that was cleaned up?
> 
> The details are empty (the first one is expanded):
> image


I'm adding a check to remove private containers from the siteData domains, thus wouldn't trigger the flag that adds it to the activity log even if there wasn't any cookies deleted.  I think that's the cause.

src/services/CleanupService.ts
kennethtran93: Resolved through 17b7bff

src/services/CleanupService.ts
kennethtran93: Resolved through 17b7bff

kennethtran93:
@sdht0 Incorporated some of the changes you mentioned.  Feel free to test it out again.  There's an active bug in regards of `use_site` with `firstparty.isolate`, which I need to work on. It will be on a new branch, that is based off of this however.

extension/global_files/browserDetect.js
```diff
@@ -6,67 +6,90 @@
  *
  * @returns {string}
  */
-var browserDetect = function() {
```
ke-d, line 9: Maybe something to look into the future, but maybe there's a npm module that can be imported to do this functionality? However, no action needed in this PR. 

My original reasoning for storing it into redux cache is so that the background and options scripts don't have to recompute the browser detect. But now I'm thinking of ways of reducing complexity for the future.

ke-d (dismissed):
Looks good to me for the new changes.

sdht0:
Thanks for the changes! Looks good to me.
One issue that I found is that notifications (3 notifications on cleanup for a domain) are being triggered even when the option is OFF in my settings.

kennethtran93:
If you're doing manual cleanup then notification will always show, per an older request. There's a new issue that wants that reverted so in the future I'll need to add another option for that.

sdht0:
Sorry I should have clarified. I meant that automatic cleanup is showing notifications and looks like a regression. Changing the option in settings to ON or OFF does not seem to have an effect, although it looks like when the settings is ON, there are a lot more notifications that are being shown.

kennethtran93:
I'll take a look soon. If the notification has a red icon then it is usually an error thrown that is always displayed. Otherwise I'll need to dig deeper into it and follow the logic.

sdht0:
Yes, looks like the notifications for normal cleanups, with individual notifications for each type. It seems to be a regression in the refactoring done in the last commit. Can you not reproduce this on your end?

kennethtran93:
@sdht0 Whoops.  Should be fixed now.  I had forgotten I had refactored some code to minimize duplication of checks but forgot to do a check from there. I've also added an option to toggle manual notifications display per other requests.  I've tested the relevant notification options and I believe it should be working properly now.

kennethtran93:
Unless there's more regressions, I promise that this will be merged upon approval this time.

src/services/CleanupService.ts
```diff
@@ -361,26 +361,33 @@ export const clearCookiesForThisDomain = async (
       );
       if (r) cookieDeletedCount += 1;
     }
-    showNotification({
+    showNotification(
```
ke-d, line 364: Hmm, it might be a good idea to decouple some of the notification logic away from Cleanup service. It can be a wrapper function around `showNotification` that takes in something that Cleanup service would return (maybe CleanReasonObject with some refactoring to include the other side data hostnames). Based on whats in that object, you can output a different types of notifications. Though this would require some refactoring that I don't expect in this PR.

ke-d (approved):

sdht0:
Thanks a lot for all the work @kennethtran93!

fabianski7:
this PR allows [FileSystem](https://developer.mozilla.org/en-US/docs/Web/API/FileSystem) to be cleaned too?

kennethtran93:
> this PR allows [FileSystem](https://developer.mozilla.org/en-US/docs/Web/API/FileSystem) to be cleaned too?

Not yet, per https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browsingData/DataTypeSet, at least in Firefox.  Seems like only Chrome/Edge Chromium has support for it at this time.

Perhaps in a future release will it be implemented.

PR #812, opened 2020-07-28
Mark ActivityLog.browsingDataCleanup as optional

ke-d:
This is to fix popup throwing errors when the user has old ActivityLogs without `browsingDataCleanup`. This PR marks that fields as optional and adds the necessary checks in the code.

### DIFF
--- a/src/services/CleanupService.ts
+++ b/src/services/CleanupService.ts
@@ -950,9 +950,9 @@ export const cleanCookiesOperation = async (
       isSafeToCleanObjects,
     );
     // Don't store domains for private browsing data
-    if (storesIdsToScrub.includes(id)) continue;
+    if (storesIdsToScrub.includes(id) || !storeResults) continue;
     for (const sd of SITEDATATYPES) {
-      if (storeResults && (storeResults[sd] || []).length > 0) {
+      if ((storeResults[sd] || []).length > 0) {
         cachedResults.siteDataCleaned = true;
         deletedSiteDataArrays[sd] = (deletedSiteDataArrays[sd] || []).concat(
           (storeResults[sd] as string[]).map((domain) => trimDot(domain)),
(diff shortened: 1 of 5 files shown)


src/services/CleanupService.ts
```diff
@@ -952,7 +952,7 @@ export const cleanCookiesOperation = async (
     // Don't store domains for private browsing data
     if (storesIdsToScrub.includes(id)) continue;
     for (const sd of SITEDATATYPES) {
-      if ((storeResults[sd] || []).length > 0) {
+      if (storeResults && (storeResults[sd] || []).length > 0) {
```
kennethtran93: While this may be less code to write, I think the check should be earlier on so as not to keep checking again?

i.e. just below line 953 (before the for loop) we could put
```javascript
if (!storeResults) continue;
```
so that the for loop is completely bypassed if that object doesn't exist.

kennethtran93 (commented):
Just a suggestion made, possibly to improve the logic workflow.

Aside from that I approve, though I forgot to change the review type :D

src/services/CleanupService.ts
ke-d: Done. As a side note, I didn't realize that that line was in a for loop until I scrolled way up. Perhaps it might time to refactor that large function in the future?

kennethtran93: I'm not sure I see much to refactor considering the inner workings aren't used anywhere else.  There's the isSafeToClean object generation (already separate), the debug sanitization steps (relies on debug pref and isSafeToCleanObject), the cookieStoreId generations (relies on cache browser), then the for loop for each container, to which it gets all cookies for each container and filters it through isSafeToClean object before cleaning cookies, and similar story in otherBrowsingDataCleanup but the entire isSafeToCleanObject is passed in for further filtering.

kennethtran93 (approved):
Thanks!

PR #816, opened 2020-07-30
feat: Summary notifications

kennethtran93:
This consolidates site data notifications.

For autoclean notifications, it has been generalized to 'Invoked cleanup of site data for <domains>'
For manual notifications, it will still show individual site data cleanup, or if 'all' was used, then all site data cleanups (minus cookies as that has its own notification) will be combined (i.e. 'Invoked cleanup of Cache, IndexedDB, ... for <domain>).

This also fixes the eslint warnings about missing return types.

A minor patch to the option browserDataCleanup object in Activity Log - Only make it required when creating new ActivityLog Object from autoclean triggers.

### DIFF
--- a/extension/_locales/en/messages.json
+++ b/extension/_locales/en/messages.json
@@ -919,6 +919,10 @@
     "message": "Show Number of Cookies for that Domain over the Icon",
     "description": "Show Number of Cookies for that Domain over the Icon"
   },
+  "siteDataText": {
+    "message": "Site Data",
+    "description": "Site Data.  Found in Activity Log Summary and Notifications."
+  },
   "sizePopupText": {
     "message": "Size of Popup (in px)",
     "description": "Size of Popup (in px).  This sets the font size in the Popup."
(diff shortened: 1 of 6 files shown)


extension/_locales/en/messages.json
```diff
@@ -919,6 +919,10 @@
     "message": "Show Number of Cookies for that Domain over the Icon",
     "description": "Show Number of Cookies for that Domain over the Icon"
   },
+  "siteDataText": {
+    "message": "Site Data",
+    "description": "Site Data.  Found in Activity Log Summary and Notifications."
```
ke-d, line 924: Looks like there's an extra space between `Data.  Found`

ke-d (approved):

extension/_locales/en/messages.json
kennethtran93: It's only a description.  During my high school computing classes, I had been instructed to use double-spacing after each period, despite never used a typewriter before.  I haven't given it much thought about changing it back to one space (including after colons), so this habit persisted over the last 10 years.  Of course this only happens with a physical keyboard.  Software keyboards can auto correct my two spaces back into one.

PR #820, opened 2020-08-02
Implement ContextualIdentityService and eventListenerActions

kennethtran93:
This implements the ContextualIdentityService, updating and removing container lists as needed (Firefox only).

Added eventListenerActions, a common function to check if event listeners are added/removed and do so accordingly.

This also adds a generic string enumeration for SettingID to be more easily used everywhere else (pre-compile that is).

Also added CookieEvents Tests.

Sep 18th added SettingService

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -26,6 +26,7 @@ import CookieEvents from './services/CookieEvents';
 import {
   cadLog,
   convertVersionToNumber,
+  eventListenerActions,
   extractMainDomain,
   getSetting,
   siteDataToBrowser,
@@ -35,6 +36,7 @@ import {
 import StoreUser from './services/StoreUser';
 import TabEvents from './services/TabEvents';
 import { ReduxAction, ReduxConstants } from './typings/ReduxConstants';
+import ContextualIdentitiesEvents from './services/ContextualIdentitiesEvents';
 
 let store: Store<State, ReduxAction>;
 let currentSettings: { [setting: string]: Setting };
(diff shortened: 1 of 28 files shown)


src/services/Libs.ts
```diff
@@ -104,6 +104,33 @@ export const convertVersionToNumber = (version?: string): number => {
...
+ * @param event The event listener
+ * @param listener The callback function to add/check/remove.
+ * @param action The EventListenerAction (add/remove).
+ */
+export const eventListenerActions = (
+  event: EvListener<any>,
+  listener: (...args: any[]) => void,
+  action: EventListenerAction,
+): void => {
+  if (!event) return;
```
ke-d: You can probably combine these two `if (!event && !event.hasListener) return;`

src/services/ContextualIdentitiesEvents.ts
```diff
@@ -0,0 +1,161 @@
...
+
+import StoreUser from './StoreUser';
+import { cacheCookieStoreIdNames, removeListUI } from '../redux/Actions';
+import contextualIdentitiesChangeInfo = browser.contextualIdentities.contextualIdentitiesChangeInfo;
+import { cadLog, eventListenerActions, getSetting } from './Libs';
+import { ReduxConstants } from '../typings/ReduxConstants';
+
+export default class ContextualIdentitiesEvents extends StoreUser {
+  public static init(): void {
+    if (!browser.contextualIdentities) return;
```
ke-d: Feel free to combine the shorter if statements like with this and `ContextualIdentitiesEvents.isInitialized`

src/background.ts
```diff
@@ -73,12 +75,17 @@ const browsingDataCleanup = async (
 const onSettingsChange = async () => {
```
ke-d: Maybe refactoring the onSettingsChange stuff to a separate file just to shorten `background.ts` a bit?

ke-d (changes requested):

src/services/Libs.ts
kennethtran93: Combined via 4153672

src/services/ContextualIdentitiesEvents.ts
kennethtran93: Combined via 4153672

src/background.ts
kennethtran93: Extraction attempted via a5cfdb7

ke-d (approved):

Issue #120, opened 2017-08-28
Search whitelist when typing


rugk:
If one types something into the input box in the whitelist settings, the add-on should search for that string in the whitelist and filter the results, so that one can see if there is already a entry for a particular website.

Also this makes the input box there a "dual-purpose" thing as one can then use it to search *and* to add entries.

mrdokenny:
I was thinking it should filter your cookie domains instead. Say you type in 'g', it would filter the cookie domains with 'g' like google.com

rugk:
What are "cookie domains"?

All domains, which once set a cookie? If so, that makes few sense, it would basically just be your browser history, i.e. you can also use your address bar for that.

mrdokenny:
Let me rephrase it. It would function more like auto-complete this way. Type in 'g', it would give a list of 'g' related cookies that you have right now. So it would be a quicker way of adding domains based on the cookies you have.

rugk:
Hmm, you could do, but then a search field for searching the whitelist would still be good.

What about doing both things. Showing a autocomplete-drop-down directly below the input (similar to how search engines do it) and still filtering the whitelist (which is even below that), so one can click anywhere to hide the auto-complete field and then have the filtered list?

kennethtran93:
Initial implementation will only search expressions/hostnames.  Think I'll leave searching for cookie-specific data to other extensions.

Closing as resolved from #941

PR #821, opened 2020-08-02
Expression enhancements

### LINKED ISSUE #814: [FEATURE REQUEST] Allow CIDR in Expression
### Is your feature request related to a problem? Please describe.
For the websites visited by IP address, whether it is the choice of IPv4 or IPv6, they are not well supported
### Describe the solution you'd like
Add CIDR support

#### ipv4 private address  
/^(?:(?:1[0-9][0-9]\.)|(?:2[0-4][0-9]\.)|(?:25[0-5]\.)|(?:[1-9][0-9]\.)|(?:[0-9]\.)){3}(?:(?:1[0-9][0-9])|(?:2[0-4][0-9])|(?:25[0-5])|(?:[1-9][0-9])|(?:[0-9]))$/

#### ipv6 private address  
I con`t find

### Additional context

### LINKED ISSUE #865: [BUG]  use of comma in regexp didn't Export and Import correctly
**Describe the bug**
Use of comma in regexp didn't Export and Import correctly.

**To Reproduce**
1. Add Greylist regexpr:   /\w\.[a-z]{2,3}$/
2. Click on Export Expressions
3. Click on Import Expressions
4. See error, it created two entries of weird expressions from  /\w\.[a-z]{2,3}$/
--1st line entry:  /\w\.[a-z]{2
--2nd line entry:  3}$/

Seems the comma between 2 and 3 in original regexp caused the issue.

**Expected behavior**

Should not be two entries on two lines, should just be one line showing:  /\w\.[a-z]{2,3}$/

CAD 3.5.1
on chrome Version 85.0.4183.83 (Official Build) (64-bit)
windows 10 Version	10.0.14393 Build 14393

kennethtran93:
- This adds the ability to search while in List of Expressions per #120.  The search algorithm is in alpha, but it should work for most people.
- This also adds the ability to use CIDR Notations per #814.

### DIFF
--- a/src/services/Libs.ts
+++ b/src/services/Libs.ts
@@ -396,36 +396,22 @@ export const getHostname = (urlToGetHostName: string | undefined): string => {
 };
 
 /**
- * Retrieves all Expressions within a single list/container.
- * @param state The WebExtension State
- * @param storeId The storeId/Container to get
- */
-export const getListSingle = (
-  state: State,
-  storeId: string,
-): ReadonlyArray<Expression> => state.lists[storeId] || [];
-
-/**
  * Returns all matched Expressions from a single list.
  * Can pass in either a single list of Expression or the entire State
  * First checks for IP, then CIDR, then falls back to Regular Expression.
  * If no input given, return all expressions from that list.
- * @param stateOrList The WebExtension State or A Single List of Expressions
+ * @param lists The Container List of Expressions from State.lists
  * @param search whether we're searching for a regex or matching
  * @param input The string for testing
  * @param storeId The storeId/Container
  */
 export const getMatchedExpressions = (
-  stateOrList: ReadonlyArray<Expression> | State,
+  lists: StoreIdToExpressionList,
+  storeId: string,
   input?: string,
-  storeId?: string,
   search = false,
 ): ReadonlyArray<Expression> => {
-  const expressions = Array.isArray(stateOrList)
-    ? (stateOrList as ReadonlyArray<Expression>)
-    : storeId
-    ? getListSingle(stateOrList as State, storeId)
-    : [];
+  const expressions = lists[storeId] || [];
   if (expressions.length === 0 || !input || input.trim().length == 0)
     return expressions;
   // Check if input is a valid IP Address (IPv4 or IPv6) (non-CIDR)
@@ -736,9 +722,9 @@ export const returnMatchedExpressionObject = (
   hostname: string,
 ): Expression | undefined => {
   return getMatchedExpressions(
-    state,
-    hostname,
+    state.lists,
     getStoreId(state, cookieStoreId),
+    hostname,
   )[0];
 };
 
--- a/src/services/Libs.ts
+++ b/src/services/Libs.ts
@@ -11,7 +11,7 @@
  * SOFTWARE.
  */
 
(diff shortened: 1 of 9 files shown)


src/services/Libs.ts
```diff
@@ -365,39 +397,96 @@ export const getHostname = (urlToGetHostName: string | undefined): string => {
...
+ * First checks for IP, then CIDR, then falls back to Regular Expression.
+ * If no input given, return all expressions from that list.
+ * @param lists The Container List of Expressions from State.lists
+ * @param search whether we're searching for a regex or matching
+ * @param input The string for testing
+ * @param storeId The storeId/Container
  */
-export const getContainerExpressionDefault = (
-  state: State,
+export const getMatchedExpressions = (
```
ke-d, line 408: This function is a little bit too long for my taste. Looks like it can be broken up into separate parts.

src/ui/settings/components/Expressions.tsx
```diff
@@ -107,16 +126,57 @@ class Expressions extends React.Component<ExpressionProps> {
...
     this.setState({
       expressionInput: '',
+      success:
+        errExps.length > 0
+          ? browser.i18n.getMessage('regexMissingEndSlash', errExps.join(' '))
+          : '',
+    });
+  }
+
+  private parseRawExpression(exp: Expression): string[] {
```
ke-d, line 150: No action needed in this PR, but I feel like we should start testing some of these React files or at least the lib functions in them, since we are starting to have a lot more logical code in the React files.

src/ui/common_components/ExpressionOptions.tsx
```diff
@@ -111,12 +112,37 @@ class ExpressionOptions extends React.Component<ExpressionOptionsProps> {
         (cookie) => cookie.domain === '' && regExp.test(cookie.path),
       );
     } else {
-      cookies = await browser.cookies.getAll(
-        returnOptionalCookieAPIAttributes(this.props.state, {
-          domain: `${trimDotAndStar(exp)}${exp.endsWith('.') ? '.' : ''}`,
-          storeId: this.toPublicStoreId(expression.storeId),
-        }),
-      );
+      try {
```
ke-d, line 117: These nested try catch blocks looks complicated, can they be broken up?

ke-d (changes requested):

src/services/Libs.ts
kennethtran93: The only part I may be able to break up is my moving the entirety of the return results from `expressions.filter` at the end as everything inside is either a first boolean return or fallback/continue parsing for other criteria, so it is harder to separate that chunk.  That filter function is 53 lines by itself (well with some comments).

src/ui/common_components/ExpressionOptions.tsx
kennethtran93: I think I resolved this through 9c92dde.  Logically this should work.

src/services/Libs.ts
kennethtran93: Should now be resolved through 533a961.

Two chunks were extracted into separate functions.  While tests still ran fine and covered, I still added some individual test cases to each new function.

ke-d (approved):

PR #822, opened 2020-08-06
Widen delayBeforeClean number input field in settings

aheckler:
The `delayBeforeClean` input field in settings was narrow enough that it could potentially hide the rightmost digits of very large numbers. For example, if you have it set to auto-clean every week (604800 seconds), the two rightmost digits are cut off:

Markup on 2020-08-06 at 16:13:39

This minor CSS change widens the field to accommodate the maximum number of digits in the value: 7. Using the maximum value of 2147483 as an example:

Markup on 2020-08-06 at 16:22:07

Also looks good in Firefox:

Markup on 2020-08-06 at 16:25:18

### DIFF
--- a/extension/settings/settings.css
+++ b/extension/settings/settings.css
@@ -21,6 +21,10 @@ input[type='number'] {
   width: 5em;
 }
 
+input[type='number']#delayBeforeClean {
+  width: 7em;
+}
+
 label:hover {
   background-color: #d3d3d3;
   cursor: pointer;


kennethtran93:
The autoclean value being that large would only work if the browser stays open for at least that long without any power interruptions or reboots.

That being said I do see the issue on hand.  While what you have contributed above may work in the meantime, though if there are any changes to the source code, especially if the input type changes or the id changes, then this would be somewhat unused (and may be forgotten altogether).  I thus propose we actually remove that css input number section (because a quick search of the src folder indicates that the input mentioned is actually the only one using it) and replace it with the relevant Bootstrap class and using the `size` attribute that will also adjust the size accordingly.

If you want to incorporate that into this PR you can, otherwise I'll (eventually) commit it on my end as I have that portion of code just tacked into whatever else I'm trying to fix before the next release.

kennethtran93 (commented):
See PR comment above.

PR #823, opened 2020-08-10
Fix FirstPartyIsolate and use_case settings.

### LINKED ISSUE #801: [BUG] Cookie counter on subdomains always zero
**Describe the bug**

The cookie counter on websites with a subdomain like `https://en.wikipedia.org/` is always zero, even when there are cookies clearly set and also deleted when leaving the sites.
Manual delete doesn't work as it shows 'No cookies found'.

On sites without a subdomain like `https://github.com/` it works fine.

**To Reproduce**

1. Go to `https://en.wikipedia.org/`
2. Open the addon menu
3. See '0 cookies'
4. Open browser console and go to Storage Inspector

**Expected behavior**

The cookies should get counted properly

**Screenshots**

image

 - OS: Ubuntu 20.04
 - Browser Info: Firefox 78.0.1
 - CookieAutoDelete Version: latest™ 3.4.0

kennethtran93:
This should fix #801.

In hindsight we should have allowed cookies stored using firstPartyIsolate to be removed even if firstPartyIsolate was (recently) disabled.  This should allow it.

~I'm having some difficulty writing test cases for these so they're not included....for the time being.~  Test Cases Added as of 50da8b4.

### DIFF
--- a/src/services/Libs.ts
+++ b/src/services/Libs.ts
@@ -160,7 +160,7 @@ export const getAllCookiesForDomain = async (
   state: State,
   tab: browser.tabs.Tab,
 ): Promise<browser.cookies.Cookie[] | undefined> => {
-  if (!tab.url) return;
+  if (!tab.url || tab.url === '') return;
   if (tab.url.startsWith('about:') || tab.url.startsWith('chrome:')) return;
   const debug = getSetting(state, 'debugMode') as boolean;
   const partialTabInfo = createPartialTabInfo(tab);
--- a/src/services/Libs.ts
+++ b/src/services/Libs.ts
@@ -465,12 +465,12 @@ export const returnOptionalCookieAPIAttributes = (
   cookieAPIAttributes: Partial<CookiePropertiesCleanup> & {
     [x: string]: any;
   },
-  firstPartyIsolate: boolean,
 ): Partial<CookiePropertiesCleanup> => {
   // Add optional firstPartyDomain attribute
+  // To fetch firstPartyIsolation cookies even if FPI is off,
+  // set firstPartyDomain to null.
   if (
     isFirefox(state.cache) &&
-    firstPartyIsolate &&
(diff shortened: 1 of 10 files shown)


kennethtran93:
/get-build

Test after recent commit for slash-dispatch upgrade.

src/services/Libs.ts
```diff
@@ -151,6 +151,134 @@ export const extractMainDomain = (domain: string): string => {
...
+      msg: 'Libs.getAllCookiesForDomain:  browser.cookies.getAll for domain.',
+      x: {
+        partialTabInfo,
+        domain: hostname,
+        firstPartyDomain: `(${proto},${mainDomain})`,
+      },
+    },
+    debug,
+  );
+  const cookies_fpi_usesite = await browser.cookies.getAll(
```
ke-d: camelCase for consistency? like `cookiesFPIUsesSite` or something

src/services/Libs.ts
```diff
@@ -151,6 +151,134 @@ export const extractMainDomain = (domain: string): string => {
...
+        msg: 'Libs.getAllCookiesForDomain:  browser.cookies.getAll for domain.',
+        x: {
+          partialTabInfo,
+          domain: hostname,
+          firstPartyDomain: `(${proto},${mainDomain},${siteURL.port})`,
+        },
+      },
+      debug,
+    );
+    const cookies_fpi_usesite_port = await browser.cookies.getAll(
```
ke-d: Same as above.

ke-d (dismissed):

src/services/Libs.ts
kennethtran93: Resolved in 
097c8cf

src/services/Libs.ts
kennethtran93: Resolved in 
097c8cf

PR #824, opened 2020-08-10
Fix:  Delay in Seconds not showing long numbers.

kennethtran93:
Alternative patch from issue raised in and will close #822.

### DIFF
--- a/extension/settings/settings.css
+++ b/extension/settings/settings.css
@@ -17,10 +17,6 @@ input[type='file'] {
   display: none;
 }
 
-input[type='number'] {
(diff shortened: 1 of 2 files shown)


extension/settings/settings.css
```diff
@@ -17,10 +17,6 @@ input[type='file'] {
   display: none;
 }
 
-input[type='number'] {
```
ke-d, line 20: Might be worth looking into the future to see how much of this css file we can put into Settings.tsx for simplification.

ke-d (dismissed):

PR #906, opened 2020-09-26
Add unlimitedStorage to ext manifest

kennethtran93:
For the future.

This will also pave the way for some form of browser sync.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -46,6 +46,7 @@
     "notifications",
     "storage",
     "tabs",
+    "unlimitedStorage",
     "<all_urls>"
   ],
   "manifest_version": 2,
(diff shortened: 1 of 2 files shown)


ke-d (commented):
We should probably wait until there is an actual code that uses this permission

kennethtran93:
> We should probably wait until there is an actual code that uses this permission

That is true.  Note that storage.local only allows ~5 MB (5,242,880 bytes).  While my current lists amount to just over 25kb, others may have more, especially if they have heavily configured it (e.g. those what do both the main domain and the wildcard subdomains at the same time) and/or have a rather large activity log (though we do truncate it to the last 10 entries).  In addition, Chrome has that hard limit, but Firefox has yet to cap it yet.

This may probably come into effect if I were to try and implement a history/internal backup function through a combination of storage.local and storage.sync (both browsers have a hard limit of ~100 kB (102,400 bytes) and cannot be increased through unlimitedStorage).

PR #910, opened 2020-09-29
Shorten ActivityLog Summary to only list 5 sites.

### LINKED ISSUE #864: [BUG] Cleanup log summary too large.
**Describe the bug**

Not actually sure if it's a bug, but it's certainly something new in 3.5.1 and it seems very unnecessary.

The issue is that for every manual clean I get a massive list of all whitelisted domains/hosts, like this:

`26/08/2020, 15:28:05 CEST - Some site data, including 0 Cookie(s), removed from: bbc.com, www.bbc.com, account.bbc.com, www.discogs.com, www.last.fm, web.archive.org ... [rest of 200 entries]`

Is this expected behaviour, and if it is, is there a setting to just have it write what it removed?

**To Reproduce**

1. Click on extension icon
2. Click on Clean
3. The huge list gets generated in the log

**Expected behavior**

I would expect it to only show the domains it clean, not the full whitelist.

**Screenshots**

n/a

 - Windows 10 Pro x64 build 2004
 - Browser Info: Vivaldi 3.3.2020.3 (Chrome/85.0.4183.80)
 - CookieAutoDelete Version: 3.5.1

**Additional context**

Might be a Vivaldi issue, I will check in older versions too.

kennethtran93:
This closes #864.

If the summary line needs rewording, feel free to comment with suggestions.

Screenshot_20200928_193837

### DIFF
--- a/extension/_locales/en/messages.json
+++ b/extension/_locales/en/messages.json
@@ -622,15 +622,19 @@
     "description": "None"
   },
   "notificationContent": {
-    "message": "Some site data, including $Num$ Cookie(s), removed from: $Websites$",
-    "description": "Some site data, including 123 Cookie(s), removed from: google.com, facebook.com",
+    "message": "Some site data, including $Num$ Cookie(s), removed from $total$ domain(s), including: $Websites$",
+    "description": "Some site data, including 123 Cookie(s), removed from 8 domain(s), including: google.com, facebook.com",
     "placeholders": {
       "num": {
         "content": "$1",
         "example": "123"
       },
-      "websites": {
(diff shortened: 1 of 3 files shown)


ke-d (dismissed):
Maybe add something like this https://css-tricks.com/line-clampin/, just to show that there is more. It can be even be as simple as adding "..." to the end or something.

kennethtran93:
I've added the ellipsis to the summary ends.  This only shows up in the notifications as well as activity log summary.

ke-d (approved):

PR #941, opened 2020-10-24
Validation for revised expressions.

### LINKED ISSUE #120: Search whitelist when typing

### LINKED ISSUE #849: [BUG]  No warning about commas or multiple domains in expressions through existing rule
**Describe the bug**

I misunderstood https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/pull/652 and moved multiple domains to the same existing rule. There was no warning that the rule was invalid.

**To Reproduce**

1. Replace an existing rule with `example.org, example.com`

**Expected behavior**

The user should be told that this rule is invalid.

 - OS: Windows 10
 - Browser Info: Firefox 79
 - CookieAutoDelete 3.5.0

### LINKED ISSUE #850: [BUG]  Wrap long expressions
**Describe the bug**

If you add a long regex to your list of expressions, it breaks the layout of the settings page and the extension popup.

**Expected behavior**
Lengthy strings should be wrapped.

**Screenshots**

wrap

**Your System Info:**

 - OS: Windows 10
 - Browser Info: Firefox 79
 - CookieAutoDelete 3.5.0

kennethtran93:
This adds validation for editing existing expressions on the Expression Table.  Resolves #849.

Note:  This is only when editing expressions, NOT applied (yet) to adding new expressions.

This only covers the more common items such as:
- Validating Regex
- Checking for Start AND End Slash for Regex
- If not Regex:
  - slash is found at beginning or end.
  - comma is found (only allowed in Regex)

If there are other validations I should do please let me know.

This also moves the reset/save buttons to the bottom of the input field and makes them larger.

image

image

Since the first commit, the following items have been added/adjusted:
- Wrapping Long expression domains through textarea to almost mimic the size of the input field.  Resolves #850.
- Shift+Enter to add to GreyList on active list.
- Validate New Expressions from manual input as well as from file import.
- Fixed crash when having a `\` at the end of a search string.

Peek 2020-10-26 22-45

### DIFF
--- a/src/ui/common_components/ExpressionTable.tsx
+++ b/src/ui/common_components/ExpressionTable.tsx
@@ -147,7 +147,7 @@ class ExpressionTable extends React.Component<
       expressionColumnTitle,
       emptyElement,
     } = this.props;
-    const { editMode, id, expressionInput, invalid } = this.state;
+    const { editMode, id, invalid } = this.state;
     const expressions =
       this.props.expressions === undefined ? [] : this.props.expressions;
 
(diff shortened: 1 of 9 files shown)


src/ui/common_components/ExpressionTable.tsx
```diff
@@ -88,6 +99,45 @@ class ExpressionTable extends React.Component<
...
+    this.editInput = undefined;
+  }
+
+  public validateEdit(): boolean {
+    if (!this.state.editMode || !this.editInput || !this.state.id) return false;
+    const inputTrim = this.state.expressionInput.trim();
+    if (!inputTrim) {
+      return this.setInvalid(browser.i18n.getMessage('inputErrorEmpty'));
+    }
+    if (inputTrim.startsWith('/') && inputTrim.endsWith('/')) {
```
ke-d: Hmm this seems like something you should probably unit test to verify the results.

ke-d (changes requested):

src/ui/common_components/ExpressionTable.tsx
kennethtran93: Moved to a function in LIbs to allow validation from List of Expressions (new input) as well.

ke-d (approved):

PR #971, opened 2020-11-20
Update README.md

zero77:
Related to:
https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/discussions/949

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,19 @@ Control your cookies! This extension is inspired by [Self-Destructing Cookies](h
 - Easily see the number of cookies for a site
 - Support for Container Tabs (Firefox 53+ only)
 
+| Cleanable Data       | Support     | Reference     |Known Issues     |
+| :------------- | :----------: | :-----------: |:-----------: |
+|  Cookies | Yes   |     |
+| Local Storage   | Yes |  [See Supported Browser Versions](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/wiki/Documentation#enable-localstorage-support)  | |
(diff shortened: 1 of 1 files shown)


zero77:
Have i missed anything.

kennethtran93:
Great start.  Perhaps add in a column for Firefox users i.e. `Per Container (contextual identities)` and could put yes to relevant ones with an asterisk `*` to link to a note regarding cross-container issues for non-cookie items?

zero77:
> Great start. Perhaps add in a column for Firefox users i.e. Per Container (contextual identities).

Would contextual identities, be ok as a column header.

> Could put yes to relevant ones with an asterisk * to link to a note regarding cross-container issues for non-cookie items?

I was expecting that any related issues would be added to Known Issues.

Also, does this mean some of the Cleanable Data types are not out of scope and being considered.
If so i will change "Out Of Project Scope" to something like under revue.

kennethtran93:
> Would contextual identities, be ok as a column header.

Probably Containers would be better as it is more of a user term, while `contextualIdentites` is more of a backend/coder term.

> 
> > Could put yes to relevant ones with an asterisk * to link to a note regarding cross-container issues for non-cookie items?
> 
> I was expecting that any related issues would be added to Known Issues.

Good point.  Yes do that then.  It is more of a browser limitation at this time than from CAD end.

> Also, does this mean some of the Cleanable Data types are not out of scope and being considered.
> If so i will change "Out Of Project Scope" to something like under revue.

I suppose.  You can put them all as **not implemented until cleanable by hostname** or something to that effect.

PR #996, opened 2020-12-16
Fix issues with partial regex domains.

kennethtran93:
While attempting to fix the internal issues raised through CodeQL, I was able to simplify the checks and replacements a bit.  This should close all the internal CodeQL issues.


ke-d (dismissed):
I don't see any new tests here. If you fixed an issue with the function then it's probably best to have new tests that cover it. Otherwise, approved.

PR #1009, opened 2020-12-24
Popup clean UI Enhancements

### LINKED ISSUE #803: [BUG]  flashing/shaking when "clean" drop down menu opened from overflow menu.
**Describe the bug**

When the menu is opened from the overflow menu in Firefox, opening the drop-down menu titled "clean" causes everything to shake back and forth quickly.

I assume this is caused by the drop-down being to large, or maybe the scrollbar changing the width of the menu.

**To Reproduce**

Steps to reproduce the behaviour:
Using Firefox for desktop:
1. Move the "Cookie AutoDelete" icon into the "overflow" menu
Image
2. Click on the icon in the overflow menu to open the menu
3. Click on the arrow at the right of the button labelled "Clean"
4. The menu should start flashing quickly

**Expected behaviour**

The drop-down should be displayed without any shaking

**Screenshots**

[Recording](https://benjiefiftysix.github.io/issue-images/Screenrecording-2020-07-16_13.31.39.mp4)

**Your System Info:**

 - OS: Manjaro KDE
 - GTK theme: Breath-Dark (controls scrollbars)
 - Browser Info: Firefox 78.0.2
 - CookieAutoDelete Version 3.4.0

kennethtran93:
This should fix the occasional flashing popup of the additional cleanup options in certain situations.  Mostly applies to Firefox's Overflow menu.

This closes #803

### DIFF
--- a/src/ui/popup/App.tsx
+++ b/src/ui/popup/App.tsx
@@ -11,7 +11,6 @@
  * SOFTWARE.
  */
 import * as React from 'react';
-import { findDOMNode } from 'react-dom';
 import { connect } from 'react-redux';
 import { Dispatch } from 'redux';
 import {
@@ -22,7 +21,6 @@ import {
 import {
   clearCookiesForThisDomain,
   clearLocalStorageForThisDomain,
-  clearSiteDataForThisDomain,
 } from '../../services/CleanupService';
 import {
   CADCOOKIENAME,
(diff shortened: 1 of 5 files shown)


src/ui/popup/App.tsx
```diff
@@ -128,7 +128,11 @@ class App extends React.Component<PopupAppComponentProps, InitialState> {
   ): Partial<HTMLButtonElement> {
     return (
       <button
```
ke-d: Perhaps it's best to move this group out to a separate file to keep the main App.tsx file cleaner.

ke-d (dismissed):

src/ui/popup/App.tsx
kennethtran93: We would need to then make some of the functions inside the App importable then as the click function and animate are also used outside of this HTML function

src/ui/popup/App.tsx
```diff
@@ -421,6 +305,77 @@ class App extends React.Component<PopupAppComponentProps, InitialState> {
             text={browser.i18n.getMessage('preferencesText')}
           />
         </div>
+        <div
```
ke-d: It's a start, but what I meant was this whole entire UI portion (308-378) should be separated into its own file to make the main file smaller.

kennethtran93: Resovled through d02a4b6

ke-d (approved):

Issue #595, opened 2019-12-04
CAD Enhancement/bugfix list


kennethtran93:
Reserving this issue for checking off what I have/will be doing. 

Feel free to subscribe to notifications on this issue for updates.

List of items that I plan to work upon soon, at the very least items that I could potentially work on more easily than others.  Just got my desktop linux refreshed after a while, so will need to get my development workflow back up again.

This list is not in order of what I will work on first.

- [x] Update dependencies (long overdue)
  - Released as v3.0.3
- [x] Automation, test, compiling (Travis CI stuff).
  - notes:  https://github.com/mozilla/sign-addon & https://github.com/mozilla/sign-addon/blob/master/.travis.yml
  - CI has transitioned from Travis to GitHub Actions.
  - PRs have been included in GitHub Actions test workflows.
  - Automated upload to Firefox and Chrome will begin with 3.2.0 (if the action workflow actually works).
- [ ] Export of small debugging info for copy-paste into GitHub Issues
  - CAD Version
  - Browser
  - Operating System (?)
  - 'Core' Settings
- [ ] Export of cleanup logs (maybe for debugging assistance?)
- [ ] Better import/export system (for multi-platform)
- [x] #567 Multiple Expression in one line ~/ reject comma-separated values...~
  - Fixed as of PR #652, released as of 3.1.1
- [x] #539 IP Regex Bug
  - Fixed as of PR #619, released as 3.0.4
- [x] #523 Chrome Localstorage
  - Initial Commit and Fix for Firefox too.  Will be released as 3.2.0
- [x] #521 white/grey from logs
- [ ] #527 Import via text file
- [x] #429 Export of 'Core' Settings
  - Added as of 3.3.0
- [x] #344 Purge expired cookies regardless of white/grey list (maybe).

kennethtran93:
Happy New Year.  I have recently received publishing permissions on all official channels and will push out a much needed dependency upgrade first before addressing anything else (It may fall under pre-release in Github first as to test out after getting it approved on Chrome and Firefox).

kennethtran93:
#615 is the final change for V3.0.3 for the time being.  

It has been officially released on Firefox on Jan 30, and Google Chrome  on Feb 3rd.

swg0101:
Just got 3.0.3 on Chrome - now just waiting for localStorage support. Thanks! :)

kennethtran93:
[...]

PR #1069, opened 2021-03-08
Add debug info to about page for easier copy-paste to GitHub.

kennethtran93:
This adds a section under 'About' page with some of the debugging information, including OS, partial browser info, and the CAD version.  It should be formatted in a way that one can copy-paste it to GitHub.

A shorter version of the settings (in the order shown in settings) is also created for copy-paste.

image

One more item off the long-overdue checklist at #595.

Signed-off-by: Kenneth T <kennethtran93@users.noreply.github.com>

### DIFF
--- a/src/ui/settings/components/About.tsx
+++ b/src/ui/settings/components/About.tsx
@@ -11,84 +11,287 @@
  * SOFTWARE.
  */
 import * as React from 'react';
+import { connect } from 'react-redux';
+import { cadLog, isFirefox } from '../../../services/Libs';
+import IconButton from '../../common_components/IconButton';
+
+const styles = {
+  buttonStyle: {
+    height: 'max-content',
+    padding: '0.75em',
+    width: 'max-content',
+  },
+};
 interface OwnProps {
   style?: React.CSSProperties;
 }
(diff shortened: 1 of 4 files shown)


src/ui/settings/components/About.tsx
```diff
@@ -11,84 +11,287 @@
...
+    });
+    return (
+      <div style={style}>
+        <h1>{browser.i18n.getMessage('aboutText')}</h1>
+        <h5>
+          {browser.i18n.getMessage('versionNumberText', ['CAD'])}:
+          <br />
+          <b>{browser.runtime.getManifest().version}</b>
+        </h5>
+        <a href="https://github.com/mrdokenny/Cookie-AutoDelete/issues">
```
ke-d: Link should probably be `https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/issues`

ke-d (dismissed):

src/ui/settings/components/About.tsx
kennethtran93: fixed in 0aa3967.  the old one still redirects to it anyway.

PR #1153, opened 2021-05-25
Dependency Updates.

kennethtran93:
Bunch of dependency updates.
Signed-off-by: Kenneth T <6724477+kennethtran93@users.noreply.github.com>


ke-d (approved):
Optional but maybe check if some of the global_files could be done via npm and webpack.

kennethtran93:
Bootstrap, jquery, and webextension-polyfill is already copied from npm through webpack, but never excluded from git commits.

PR #1205, opened 2021-08-09
Language Clarifications

### LINKED ISSUE #754: Rename whitelist to something
In the face of all the current things going on in the world it would be a nice way of showing support by renaming graylist and whitelist to something else. You for sure will be able to give that concept another name that might not carry the co notation that the term whitelist carries with it.

Thanks ❤️ .

kennethtran93:
Resolves #754.

If anybody finds a place where the existing terms did not migrate after this version, please let me know.


ke-d (dismissed):
Hmm I'm not sure if it's necessary to have it to have Whitelist, greylist, etc. be a changeable option. Might as well just change it to Keep list, Restart list, etc. and then note the changes in the documentation. It will probably make the PR less complicated also.

kennethtran93:
> Hmm I'm not sure if it's necessary to have it to have Whitelist, greylist, etc. be a changeable option. Might as well just change it to Keep list, Restart list, etc. and then note the changes in the documentation. It will probably make the PR less complicated also.

The issue referenced in question has seen some debate, so I'd rather just be able to provide both options.  This way those that want to change can do so, and those that would like to continue using the existing terms can do so without making any changes as well.

While the PR may look like a bit much, here's what it boils down to:
- Some accessibility enhancements to settings page (still needs work but part of it is there)
- Extracted the editable text field from the popup/expression table to be able to use in setting area as well.
- Added function to check if word was changed, otherwise provide default list name/type.
- Localization changes to reflect changed words.
- Whatever Prettier reformatted again.

neroux:
> This way those that want to change can do so, and those that would like to continue using the existing terms can do so without making any changes as well.

While I fully understand where you are coming from and what you are trying to achieve, I am afraid it still is caving in to unreasonable demands, which are solely based on some political agendas and follow some fashion trend.

What if someone finds offence with _cookie_ next? Is that going to be "customisable" as well? Will every string in the extension be eventually changeable? Where do we draw the line?

kennethtran93:
> While I fully understand where you are coming from and what you are trying to achieve, I am afraid it still is caving in to unreasonable demands, which are solely based on some political agendas and follow some fashion trend.

I don't think this is an unreasonable demand.  Sometimes change is for the better, along with an open mind.

> What if someone finds offence with _cookie_ next? Is that going to be "customisable" as well? Will every string in the extension be eventually changeable? Where do we draw the line?

I think I already answered this, but if that's the case, then the action we'll take will depend on the the changes that the web and internet browsers do as well.

neroux:
> I don't think this is an unreasonable demand. Sometimes change is for the better, along with an open mind.

That's the very point, there's no improvement or _change for the better_.

> if that's the case, then the action we'll take will depend on the the changes that the web and internet browsers do as well.

I am not aware of any changes implemented by browsers. Some services gave in to peer pressure and adjusted strings for no reason.

It's a very vocal minority who creates trouble here. The linked issue is the best example, the overwhelming majority voted against such a change, nonetheless it was implemented, when there are probably more pressing issues at hand.

kennethtran93:
Then we'll call it improvement and clarity of the language.  If the younger/future generation doesn't already understand white and black in the tech context then at least allow and restart should be more understandable without looking them up.

neroux:
I am afraid that is a false pretence. There's no "clarity" necessary here in the first place. Aforementioned example cookie is equally misleading, there's nowhere a biscuit or cookie in sight when it comes to that unique identifier value in the header. But no changes planned here as I understand.

Yes, it is your extension and you are obviously free to name things the way you see fit, but you do ask for donations and - while I do not pretend that I am current contributor or had immediate plans to become one - I am afraid I cannot say taking such a political stance will sway me very much to consider it in the future.

As mentioned, about 80% were against that change and it was still implemented (as optional) even though - just to put it into perspective - there are five current bugs and about 40 other proper feature requests.

I apologise, but this is _caving in_ and catering to aforementioned vocal minority with their agenda.

AderitoSilva:
I think a good name for the lists would be “**Never**” (for whitelist), “**On Tab Closed”** (for blacklist) and “**On Session Ended**” (for greylist). In my perspective, that would serve the main purpose of self-explaining when the cookie automatic deletion occurs and, as a bonus, would even put an end to the black/white lists debate. On that referred issue, I explain some more details about this suggestion.

Not that I'm adhering to that debate, but I think the _list_ concept should really be changed to something else, as it seems a bit confusing, at least to me. It seems confusing because the purpose of the lists is not clear enough, specially the greylist. And, the _events'_ concept seems to me as a nice alternative, as it would self explain the purpose of each list. Later on, more events could be added, like an “On Domain Exited” event, which deletes cookies after we navigate out of a domain within a tab, considering the appropriate auto-deletion delay. I imagine such a change would require just visual string changes.

I suggest the option to set a visual name for each list to not be included, because it adds additional complexity into the extension settings, which doesn't benefit most users at all. In my perspective, the decision about what technical terms to use should not be postponed to users, because it consumes more of their own effort, without providing them with any additional benefit.

Regarding that specific debate, it seems to me as a fight where there are no winning sides — to please one side, you have to go against the other, and vice-versa. My experience taught me that, sometimes, the only way to not lose is to not play. I strongly recommend you to try to keep CAD development out of this kind of confrontations, as it serves a purpose that is ethically technological, but doesn't benefit from being conditioned by cultural reasons (without rational/logical motives). Sometimes, efficiency and culture go in opposite directions, where culture changes naturally while efficiency requires cognizant research. This is just a mere suggestion, based on my own personal experiences.

Keep up the good work, and thanks for this great extension! :)

kennethtran93:
This PR has been updated to just be Language Clarification.

src/background.ts
```diff
@@ -326,10 +328,45 @@ browser.runtime.onInstalled.addListener(async (details) => {
           }
         }
       }
-      if (convertVersionToNumber(details.previousVersion) < 300) {
-        store.dispatch({
-          type: ReduxConstants.RESET_COOKIE_DELETED_COUNTER,
-        });
+      // Language Migration 3.7.0+
+      if (convertVersionToNumber(details.previousVersion) < 370) {
```
ke-d, line 332: I recommend adding a test for this if you haven't already (hard to tell in this large PR), just to verify the behavior from one state to another. Another option is to split the setting migration + test into a separate PR.

src/redux/Actions.ts
```diff
@@ -232,12 +192,9 @@ export const resetAll = (): RESET_ALL => ({
 });
 
 // Validates the setting object and adds missing settings if it doesn't already exist in the initialState
-export const validateSettings: ActionCreator<ThunkAction<
-  void,
-  State,
-  null,
-  ReduxAction
->> = () => (dispatch, getState) => {
+export const validateSettings: ActionCreator<
```
ke-d, line 195: I recommend next time having a PR just for prettifying the code and having a PR check for Prettier. It will also make future diffs a lot cleaner. Right now, it's hard to tell what actually changed.

ke-d (changes requested):

PR #1410, opened 2022-06-16
Initial Test of browser type detection for Firefox

### LINKED ISSUE #1409: [BUG] Today Firefox Nightly + your ext 3.6.0 and 3.8.0 = ERROR
### Acknowledgements

- [X] I acknowledge that I have read the above items

### Describe the bug

After today update Firefox Nightly error appears =
"Error! 
Type error for parameter options (Unexpected property "origins") for browsingData.remove.
"
The error appears periodically - when the browser is launched, when the pages are opened.
I restored the old version, 3.6.0, the error still occurs.

If you turn off your expansion = error stops appearing.

Apparently, the developers of Firefox changed something again.

### To Reproduce

1. update firefox nightly
2. firefox nightly start - error appears

### Expected Behavior

no error

### Screenshots

_No response_

### System Info - Operating System (OS)

win 10 64

### System Info - Browser Info

firefox nightly today update (everyday new)

### System Info - CookieAutoDelete Version

3.8.0 and 3.6.0

### Additional Context

Now it’s time to fix ext, and not before.

kennethtran93:
In the recent versions of Firefox, the existing function/method/object that we have been using to detect if the browser was Firefox has been deprecated and would be removed starting in 103.  This should be a sufficient patch fix for now until other browsers actually implement `browser.contextualIdentities`.

If anybody else know of another (possibly better) way of determining if the browser was Firefox (and Firefox Android) or not, do let me know (including the method).  Checking the User-Agent string (or any functions/libraries that would ultimately rely on this) is NOT one of them as these can be spoofed/changed by the user.

Fixes #1409.

Signed-off-by: Kenneth T <6724477+kennethtran93@users.noreply.github.com>

### DIFF
--- a/extension/global_files/browserDetect.js
+++ b/extension/global_files/browserDetect.js
@@ -19,8 +19,8 @@ var browserDetect = function () {
     !!window.opera ||
     navigator.userAgent.indexOf(' OPR/') >= 0;
 
-  // Firefox 1.0+
-  var isFirefox = typeof InstallTrigger !== 'undefined';
+  // Firefox 1.0+ (InstallTrigger only available up to FF102.  Webextension API browser.contextualIdentities still only available in Firefox/Firefox Android so use that for now.)
+  var isFirefox = typeof InstallTrigger !== 'undefined' || typeof browser.contextualIdentities !== 'undefined';
 
   // Safari 3.0+ "[object HTMLElementConstructor]"
   var isSafari =
--- a/src/ui/settings/ReleaseNotes.json
+++ b/src/ui/settings/ReleaseNotes.json
@@ -1,6 +1,12 @@
 {
   "releases": [
     {
+      "version": "3.8.1 - unreleased",
+      "notes": [
+        "Fixed:  Browser detection in Firefox 103+ as the previous browser feature has been deprecated and removed. Fixes #1409 via PR #1410."
+      ]
+    },
+    {
       "version": "3.8.0",
       "notes": [
         "Added:  Option to keep or clean existing site data on new enables.",


gilic:
*This user had previously requested download links for build 2507168728 that have since expired.*

kennethtran93:
Re-testing some workflow action. Ignore upcoming bot posts.

CyberFreakUK:
*This user had previously requested download links for build 2507168728 that have since expired.*

wagnerand:
Firefox Add-ons team member here. Suggestion from our webextension engineers:

Use `browser` if you are not using a polyfill, or https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBrowserInfo.

wagnerand:
Firefox Add-ons team member here. Suggestion from our webextension engineers:

Use `browser` if you are not using a polyfill, or https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBrowserInfo.

kennethtran93:
> Firefox Add-ons team member here. Suggestion from our webextension engineers:
> 
> Use `browser` if you are not using a polyfill, or https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBrowserInfo.

Thank you for your suggestions.  I am using the `webextension-polyfill` so that the code can be used in Chrome.  As a feature detection for Firefox and Firefox Android, I'll use the `browser.contextualidentites` check for now as this appears to be a Firefox-only item (unless you believe there's something better to use for feature-detection).  If in the future Chrome ever implements this in the future, then I'll look for alternate functions/objects that only Firefox/Firefox Android has.

Archer2011:
*This user had previously requested download links for build 2507168728 that have since expired.*

lucashealman:
*This user had previously requested download links for build 2600284504 that have since expired.*

PR #1628, opened 2024-06-23
Fixed outdated MDN link, i18n both AMO links

Gitoffthelawn:


README.md
```diff
@@ -1,4 +1,4 @@
-[link-amo]: https://addons.mozilla.org/en-US/firefox/addon/cookie-autodelete/
+[link-amo]: https://addons.mozilla.org/firefox/addon/cookie-autodelete/
```
kennethtran93, line 1: Doesn't appear this change would make any difference - for me the revised link would still redirect to include `en-US`, but if AMO is automatically adding it in then removing it wouldn't hurt to allow for auto localization of webpage to their own language.

README.md
```diff
@@ -8,7 +8,7 @@
 
 # Cookie AutoDelete
 
-Control your cookies! This extension is inspired by [Self-Destructing Cookies](https://addons.mozilla.org/en-US/firefox/addon/self-destructing-cookies/). When a tab closes, any cookies not being used are automatically deleted. Prevent tracking by other cookies and add only the ones you trust. Easily import and export your cookie whitelist.
+Control your cookies! This extension is inspired by [Self-Destructing Cookies](https://addons.mozilla.org/firefox/addon/self-destructing-cookies/). When a tab closes, any cookies not being used are automatically deleted. Prevent tracking by other cookies and add only the ones you trust. Easily import and export your cookie whitelist.
```
kennethtran93, line 11: Should probably remove the link entirely at this point since that extension is no longer visible on the site at all.

kennethtran93 (approved):
I'll approve as-is.

README.md
Gitoffthelawn: > Doesn't appear this change would make any difference - for me the revised link would still redirect to include `en-US`, but if AMO is automatically adding it in then removing it wouldn't hurt to allow for auto localization of webpage to their own language.

Yes, AMO auto-localizes for the user's locale.  Thus it's best practice to not force a locale.

README.md
Gitoffthelawn: > Should probably remove the link entirely at this point since that extension is no longer visible on the site at all.

Yes, can remove link or use Wayback Machine.  Wayback is still offline though. :(

README.md
krystian3w: For Firefox 123+ at Android could be shorter:

https://addons.mozilla.org/firefox/addon/cookie-autodelete/

Firefox less 123 may need use link for PC/macOS with "/firefox".

PR #1735, opened 2025-07-27
Migrate to MV3 and Modern Redux

median-dxz:
**Note: This PR is currently undergoing beta testing**
 
This will close issues: #1655 #1634
 
So, this is quite a big PR. I've already outlined my changes in [#10](https://github.com/median-dxz/Cookie-AutoDelete-MV3/issues/10), and there are explanations in the commit messages as well. Here, I just want to clarify some of the most important points:
 
---
 
**Q: Why is this PR so large?**  
A: Because I migrated **all the dependencies** to their latest major versions, which introduced a lot of breaking changes, and I had to update the corresponding test suites. These changes cover build tooling, state management, UI styles, TypeScript types, and more.
 
Most of the changes focus on organizing imports, refining TypeScript types, and adapting unit tests. The other major changes include modifications to configuration files and a refactoring of Redux logic. Since the latest version of `@reduxjs/toolkit` is being used, this refactoring is necessary.

---
 
**Q: What does the MV3 migration include?**  
A: In summary, the main changes are:
1. Changed background scripts to Service Workers and modified `browserDetect` so it can run in the SW context
2. Migrated from `browserAction` to `action`
3. Fields change in `manifest.json`
4. Replaced `tabs.executeScript` with `scripting.executeScript`
5. DX: Migrated all TS types to `@types/webextension-polyfill`
6. Fixed: Ensure the timeout is now always triggered in the Service Worker environment
7. Fixed: Fix duplicate menu creation errors
8. Updated: Register event listeners synchronously in the background script

Also, a small reminder: the extension does not actually use `webRequest`, so there is no need to deal with it.
 
 Refer to: [1634#comment](https://github.com/Cookie-AutoDelete/Cookie-AutoDelete/issues/1634#issuecomment-2452797529)

---
 
**Q: Has the extension's business logic changed? Is the configuration still compatible?**  
A: Other than not using `tabs.create` to open the options page anymore (which means the config page will now default to the welcome screen), the business logic remains unchanged and the configuration is fully compatible.
 
---
 
**Q: Have the unit tests passed?**  
A: Yes, they have.
 
---
 
**Q: What about user testing?**  
A: I have tested it myself on the latest version of Chrome and found no issues so far. I haven't tested it on Firefox or other extension environments yet.
 
---
 
If you have any questions, feel free to discuss them with me in this PR!

PR #1764, opened 2025-09-09
upgrades

alaricd:


kennethtran93:
Would most of these upgrades be already done in #1735?

PR #1803, opened 2025-12-26
Let Firefox on Android manage all storage types.

mlindsay:
Firefox on Android now supports more than just cookies so the extension should allow users to manage those as well.

Replaces many uses of isFirefoxNotAndroid with simply isFirefox, but leaves in place some that are mainly about UI limitations. Completely removes the isFirefoxAndroid helper.

Without this many people may not realize that this extension is letting local storage persist through cookie cleaning, especially if they are familiar with the extension on desktop.

### DIFF
--- a/__tests__/services/Libs.spec.ts
+++ b/__tests__/services/Libs.spec.ts
@@ -43,6 +43,7 @@ import {
   returnOptionalCookieAPIAttributes,
   showNotification,
   sleep,
+  supportedStorageTypes,
   throwErrorNotification,
   trimDot,
   undefinedIsTrue,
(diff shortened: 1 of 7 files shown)


kennethtran93:
Instead of fully removing it, it needs to check if the extension is installed in version 85+ for Firefox Android, since earlier versions do not have these capabilities.

mlindsay:
Hey Kenneth,

Just wanted to verify that (Firefox && android && ffVersion >= 85) is definitely what you want before I try to update the change. In part because the capability was added in Firefox 58 and I'm not sure if you simply typo'd that. I understand release 85 is when Firefox Android started allowing most normal extensions back onto it, but I don't think it is when the capabilities were actually added. So overall I don't think there was ever a path for Firefox Android to load this extension without using developer escape hatches prior to 85, so it can't really be broken on those older versions.

Thanks

kennethtran93:
> Hey Kenneth,
> 
> Just wanted to verify that (Firefox && android && ffVersion >= 85) is definitely what you want before I try to update the change. In part because the capability was added in Firefox 58 and I'm not sure if you simply typo'd that. I understand release 85 is when Firefox Android started allowing most normal extensions back onto it, but I don't think it is when the capabilities were actually added. So overall I don't think there was ever a path for Firefox Android to load this extension without using developer escape hatches prior to 85, so it can't really be broken on those older versions.
> 
> Thanks

I'm relying on the Compatibility table from MDN (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browsingData/DataTypeSet#browser_compatibility).  If I recall there was a Firefox engine/branch change of sorts most likely around version 79 or so (or at least when it was on the public release around 2020.  I know earlier versions of Firefox Android had limited site data cleanup (primarily cookies) before 85.  Take the example of indexedDB/localstorage - according to the table linked above it was only added/available since 85+.

mlindsay:
Thanks for the link and the explanation. I've updated the change to follow with that chart. Glad for your testing as well, I made a mistake in transcribing the version numbers in one place and it triggered one of tests to fail.

mlindsay:
Just nudging for review. The change has been working well on my phone on Firefox Nightly.

mlindsay:
Hey Kenneth. Please have a look at this. I would really like it to work properly on mobile. I've been relying on my build here for a while now without issue, but it requires that I run Firefox nightly in order to allow unsigned extensions.

Thanks

